PR #48, opened 2026-07-08
fix: profile workflow QA — list refresh, honest deletes, primary badge, TV manage mode, duplicate names

RXWatcher:
## Problem

Four issues from the 2026-07-08 phone QA pass on the profile workflow, plus what the investigation turned up on TV:

1. **Created profiles didn't appear** until leaving and re-entering the picker — the mobile VM only loads in `init` and survives the push/pop to Create/Edit. (TV already had the ON_RESUME reload; mobile never got it.)
2. **Deleting the signed-in profile claimed it would delete watch history, then silently did nothing.** Two stacked causes: the previous fix cleared local profile credentials *before* the DELETE, but those very headers authorize profile management server-side, so the delete always 403'd (and the local selection was wiped anyway); and the error the VM set was immediately nulled by the follow-up `loadProfiles()`, so the user saw nothing. Additionally the server never deletes the *primary* profile (409 `primary_profile_protected`), so the dialog promised something impossible.
3. **No indicator for the primary profile.**
4. **Unlimited identically-named profiles** — no validation anywhere in the stack.
5. (Found along the way) **TV's profile manage mode was dead code** — the pencil/delete UI and VM state existed, but nothing ever rendered tiles with `manageMode = true`, so edit/delete were unreachable on TV. TV also still had the original "all profiles gone" bug: no credential clear after a successful signed-in delete.

## Changes (one commit per concern)

- **`fix(mobile)` picker QA** — ON_RESUME reload (TV pattern); delete-then-clear ordering with the clear still preceding the reload; server error messages surfaced and kept visible across the reload; crown badge (`WorkspacePremium`) on the primary tile and no delete affordance for it in manage mode.
- **`feat(tv)` manage mode + parity** — Manage Profiles / Done pill under the grid; managing routes tile clicks to edit and shows the delete pill on non-primary tiles; clear-after-successful-delete of the signed-in profile before the reload; server messages surfaced; crown badge. Deliberate tvOS deviation (tvOS has no Manage Profiles button) since the picker is Android TV's only profile-management entry point — the screen source test now pins the pill instead of forbidding it.
- **`feat(profiles)` duplicate names** — client half of Silo-Server/silo-server#342: shared `hasProfileNamed` (trimmed, case-insensitive, excludes self on rename) pre-checks in create/edit on both platforms, and those forms now surface server messages (`name_conflict`, `profile_limit_reached`) instead of generic failures. Works against older servers too, where the pre-check is the only guard.

## Testing

- `:androidApp:assembleDebug`, `:androidTvApp:assembleDebug` build clean.
- Full unit-test suite passes (includes new `ProfileNameConflictTest` for the shared helper and the updated TV picker source test).
- Server-side counterpart merged separately as Silo-Server/silo-server#342; store-level atomicity follow-up tracked in Silo-Server/silo-server#343.

## Risks / follow-up

- The Manage Profiles pill is a visible TV UI change; if strict tvOS parity is preferred, management would need a different entry point (e.g. Settings).
- Non-primary signed-in profiles attempting manage actions get the server's 403 explanation verbatim; iOS's "borrow primary context with PIN" flow is not ported.

## AI-use disclosure

Implemented with Claude Code (Claude Fable 5); build- and test-verified locally.

### DIFF
--- a/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/profiles/CreateProfileViewModel.kt
+++ b/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/profiles/CreateProfileViewModel.kt
@@ -3,7 +3,9 @@ package org.siloserver.silo.android.ui.screens.profiles
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import org.siloserver.silo.model.profile.CreateProfileRequest
+import org.siloserver.silo.model.profile.Profile
 import org.siloserver.silo.model.profile.canonicalProfileQualityPreference
+import org.siloserver.silo.model.profile.hasProfileNamed
 import org.siloserver.silo.network.ApiResult
 import org.siloserver.silo.repository.ProfileRepository
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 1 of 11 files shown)


RXWatcher:
@coderabbitai review

PR #51, opened 2026-07-09
fix(profiles): phone profile edit can clear PIN and turn subtitles off

RXWatcher:
The phone `EditProfileViewModel` carried the same `explicitNulls=false` trap that PR #50 fixed on the TV app but never ported to phone. Two edit-flow bugs:

- **Can't clear a PIN.** Disabling a PIN sent `pin = null`, which SiloJson omits from the PUT, so the server kept the old PIN. Now tracks `loadedHasPin` and sends an explicit `""` to clear (server treats `""` as clear, `null`/omitted as unchanged).
- **Can't turn subtitles off.** Selecting subtitle mode "Off" mapped to `null`, likewise omitted, so the previous mode stuck. Now sends the literal `"off"`.

Both are straight ports of `TvEditProfileViewModel`'s already-correct handling (the `loadedHasPin → ""` PIN mapping and the `"off"` subtitle mapping). Create and delete flows were already correct on both platforms and are untouched.

Gated on `:androidApp:assembleDebug` + `./gradlew test` (debug + release unit suites) — all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed profile updates so PIN changes are saved correctly when enabling, disabling, or leaving a PIN unchanged.
  * Improved subtitle settings so turning subtitles off now applies reliably instead of being treated as an unset value.

### DIFF
--- a/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/profiles/EditProfileViewModel.kt
+++ b/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/profiles/EditProfileViewModel.kt
@@ -43,6 +43,14 @@ class EditProfileViewModel(
     private var existingProfiles: List<Profile> = emptyList()
 
     /**
+     * Whether the loaded profile already had a PIN. Needed to distinguish
+     * "remove the existing PIN" (send an explicit empty string) from "there was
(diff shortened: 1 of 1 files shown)


RXWatcher:
Superseded by #52, which folds this profile PIN/subtitle fix (as commit `06cb58f1`) into the consolidated phone + TV QA sweep. Closing in favor of the single PR.

PR #60, opened 2026-07-11
Phone swipe-to-seek: live feedback + inner-80% safe-box

RXWatcher:
Jim's phone (Fold) feedback on the video player's horizontal swipe-to-seek:
> The slide to rewind/forward gives no feedback. Idk if it's rewinding or forwarding and how much. Also a safe box would be nice so the feature only activates in the inner ~80% of the screen. Sometimes I just want to swipe back not rewind.

## Changes (`PlayerGestureHandler.kt`, phone only — TV has no swipe)
1. **Live scrub feedback** — the horizontal seek drag previously only applied on release with no on-screen indication. Added a live HUD centered on screen: a rewind/forward glyph, the signed delta from the drag's start (`+1:23` / `−0:45`), and the target time. Tracks the finger during the drag, vanishes on release.
2. **Inner-80% safe-box** — the seek fired across the full width, so an edge swipe scrubbed instead of navigating back. Now a seek only starts when the drag begins in the inner ~80% (`SeekEdgeExclusionFraction = 0.1` each side); a drag starting in the outer band is left **unconsumed** so it can fall through to a system back-swipe.

## Needs on-device verification
Whether an outer-band drag actually triggers the **system back gesture** depends on the device's gesture-nav zone — the unconsumed passthrough is the right approach in-code, but it should be confirmed on the Fold. The scrub HUD + the "outer band no longer scrubs" behavior are deterministic.

## Verification
- `./gradlew :androidApp:assembleDebug` — BUILD SUCCESSFUL.
- UI/gesture change; not yet felt on-device.

## Summary by CodeRabbit

* **New Features**
  * Added an on-screen seek preview showing rewind/fast-forward direction, time adjustment, and target playback time while scrubbing.
  * Improved horizontal seeking with a safe area that prevents accidental conflicts with system back-swipe gestures.

* **Bug Fixes**
  * Horizontal drags beginning near the player’s outer edges no longer trigger unintended seeking.

### DIFF
--- a/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/player/PlayerGestureHandler.kt
+++ b/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/player/PlayerGestureHandler.kt
@@ -9,10 +9,12 @@ import androidx.compose.animation.core.tween
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
 import androidx.compose.foundation.background
-import androidx.compose.foundation.gestures.detectHorizontalDragGestures
+import androidx.compose.foundation.gestures.awaitFirstDown
+import androidx.compose.foundation.gestures.awaitHorizontalTouchSlopOrCancellation
 import androidx.compose.foundation.gestures.detectTapGestures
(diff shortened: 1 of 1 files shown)


RXWatcher:
@coderabbitai review

PR #74, opened 2026-07-16
fix(audiobook): align chapters across alternate parts

RXWatcher:
## Summary
- select one consistent audiobook file version per ordered part
- prevent alternate formats from being stitched into the same whole-book timeline
- assign globally unique indexes to chapters merged across multipart audiobooks
- keep mobile and Android TV on the same shared chapter timeline

## Root cause
The server returns alternate files grouped by audiobook part. Android passed every returned file into the multipart timeline, so each alternate was treated as another sequential part. That shifted later chapter offsets and could put the chapter list off by an entire part, including cases reported as 13 chapters.

## Verification
- `./gradlew :shared:testDebugUnitTest --tests 'org.siloserver.silo.audiobook.AudiobookTimelineTest'`
- `./gradlew :androidApp:testDebugUnitTest :androidTvApp:testDebugUnitTest :androidApp:assembleDebug :androidTvApp:assembleDebug`

Both client test suites and both debug APK builds pass.


RXWatcher:
Closing this PR because its branch history included stale integration commits, which caused a misleading large diff and merge conflicts. Replacing it with a clean upstream/main-based PR containing only the audiobook chapter alignment fix.

PR #98, opened 2026-07-23
security(android): harden project trust boundaries

RXWatcher:
## Summary

Two stacks in one branch, in commit order.

**Performance and stability** (29 commits, `eaef12d5`…`49cae1ff` — formerly #97)
- bound and coalesce startup/home hydration across Android phone and TV
- pause off-route detail enrichment and isolate playback-clock recomposition
- make teardown and final-position persistence lifecycle- and identity-safe
- preserve playback state across recreation and keep TV duration on the live clock

**Security hardening** (25 commits, `af9f4a94`…`b4a43575`)
- centralize origin validation and keep credentials, refreshes, and redirects on the authenticated Silo origin
- require per-origin consent for cleartext HTTP and authenticate custom PiP actions
- isolate EPUB WebView resources, sanitize active content, bound archive/network inputs, and contain identity-scoped storage paths
- remove implicit legacy permissions and validate merged phone/TV manifests
- pin and verify Gradle, GitHub Actions, native Dolby/libdovi provenance, dependency locks, checksums, and Rust OSV results

## Stack

- ~~stacked on #97~~ — #97 was **closed without merging**, so its 29 perf commits stay here and land with this PR. This is no longer a security-only diff, and it no longer reduces to one.
- SiloCast is explicitly excluded and no SiloCast paths are changed.

## Rebase (2026-07-25)

Rebased from `428e9678` (#94) onto `5047c4d2`, picking up #99 (progressive direct-play range resume) and #100 (FCM notification metadata).

Two conflicts, both real:

- **`AuthenticatedDataSourceFactoryTest.kt`** — #99 added the ETag/If-Range resume tests with a `FakeHttpDataSource`; this branch added the subtitle-limit tests with a `ByteArrayDataSource`. Git interleaved the two fake classes. Both test groups and both fakes kept intact.
- **`gradle/libs.versions.toml`** — git auto-merged this one *cleanly into a broken file*. #99 already declares `okhttp-mockwebserver` in the media3 block; this branch declares the identical key near `junit4`. A duplicate key is a hard version-catalog parse failure, so the build died at configuration time. The branch's duplicate is dropped, folded into `fix(network): keep credentials on the silo origin` so no commit in the branch is left broken.

One new commit, `build(android): refresh locks and verification metadata for the rebase` — #99 added the media3 test-utils/robolectric stack to `:android-shared`, which the branch's lock state (generated against the old base) fails closed on. Note for anyone regenerating: the hardening plan documents metadata-then-locks, but that order only works from a clean slate. With stale locks present nothing resolves, so locks have to be written first.

## Verification (post-rebase)

- 5,330 test executions across the four modules, debug and release: `shared` 1,546, `android-shared` 1,626, `androidApp` 918, `androidTvApp` 1,240
- phone and TV debug APKs assembled
- debug and release merged-manifest permission policies passed
- `--dependency-verification=strict` passed; live Rust OSV query passed (no affected crates.io packages)
- lock additions are #99's transitive test closure only, no new repositories

**One failure, inherited from `main` and not caused by this branch:** `PlaybackApiTest."v3 start uses canonical endpoint and negotiation fields"` fails on a clean `origin/main` checkout too. #99 added `DIRECT_STREAM_RESUME_V1_FEATURE` to the client's negotiated feature list without updating the test's expected list. This branch does not touch that file. It needs a one-line fix on `main`, separately from this PR.

## Not included

The two CodeRabbit findings fixed in `c8481f1c` (failed final-position writes discarded instead of re-queued; Up Next countdown not cancelled on exit) exist only on the fork's `main`. They apply to the perf commits carried here and are **not** on this branch.

## CI

- intentionally skipped per maintainer request; verified locally


RXWatcher:
Closing as superseded by Silo-Server/silo-android#108.

This branch turned out to be an older line of work that `main` has since moved past, rather than work still missing from it. Checking commit by commit rather than by subject line:

- The leading commits cherry-pick to **empty** against current `main` — the content is already there, squashed into `feat(android): performance/stability and transactional subtitles` under different commit messages.
- `perf(tv): curb home startup fan-out` conflicts against `main`'s already-squashed version of the same change, and also touches source-grep tests that #102 removed.
- Net tree difference is +11,335 / −36,965: `main` carries ~37k lines this branch lacks, while most of the ~11k on this side is stale state (deleted tests, older versions of files that have since changed).
- Spot checks agree — e.g. `HomeViewModel` on `main` already documents the coalescing behaviour this branch introduced.

Rebuilding it would mean reconstructing which fragments are genuinely absent and reconciling each against a squashed descendant, which is archaeology rather than a rebase. #108 carries the current state of all of it.

If a startup or playback performance regression turns up that traces back to something here, I'll port that specific behaviour deliberately rather than replay the branch.

PR #101, opened 2026-07-25
fix(settings): realign device settings with the server contract

Quick104:
Android's copy of the settings contract had drifted from the server registry in `internal/api/handlers/settings.go`. Verified findings from Silo-Server/silo-server#376.

## What changed

**`PlaybackSettingsKeys.kt`**
- `NextUpPromptSeconds`: `player.next_up_prompt_seconds` → `playback.next_up_prompt_seconds`, matching the server and Apple
- Eleven keys removed from the server-sync list (`DeviceSettings`): `player.match_frame_rate`, `player.sleep_timer_default_minutes`, and nine `subtitle.*` fields. The server registers none of them, so `keyUsesDeviceScope` returns false and both the set and delete handlers 400 before validation — **every write and every reset for these was being rejected**

**`AndroidPlayerSettingsStore.kt`**
- Playback speed clamp `4.0` → `3.0`, matching `validateFloatRange("player.playback_speed", 0.25, 3.0)`
- `player.dv_profile7_hdr10_fallback` default `true` → `false`, matching the server and Apple
- `setMatchContentFrameRate` and `setSleepTimerDefaultMinutes` move to the `*Local` helpers — identical preference keys, identical clamps, read flows untouched, so **local behavior is preserved exactly**; they simply stop syncing
- New `migrateNextUpPromptKey` (see below)

**Tests** — the speed clamp assertion was **updated, not weakened**: it now asserts both bounds (`10.0 → 3.0`, `0.01 → 0.25`). Three new migration cases.

## The migration bug, found in review

The first commit shipped a **read-only** fallback to the old DataStore slot and claimed a later write or refresh would migrate it forward. Neither does, and a refresh actively destroys it.

`applyEffectiveLocally` writes the server's effective value into the canonical slot for every registered key. Before the rename that could not clobber anything — the `player.` spelling was unregistered, so the server returned an empty value and the Int branch skipped the write. The canonical key **is** registered, with a registry default of `30`, so the first refresh after upgrade writes 30 and it shadows the legacy slot permanently. `refreshFromServer` runs on settings-screen open, player load, TV detail, and `ServerDrivenConfigRefresher` — within seconds of upgrading. Affected users are exactly those whose value never reached the server: a pre-alias server, or a flush that failed silently.

`migrateNextUpPromptKey` copies the legacy slot forward once per scope, removes it, and **enqueues the migrated value** so the server stops reporting the registry default. Copying locally is not enough on its own — the refresh would still overwrite it. It runs at the top of `ensureMigrated`, which every read path and every write path funnels through, and carries its own sentinel because existing installs have already recorded the `ensureMigrated` one; folding it in there would skip precisely the users who need it.

Suppressing the refresh write instead (skipping keys the server reports without a device override) was considered and **rejected**: the subtitle-appearance block deliberately propagates a server-side reset, and a blanket guard would break that for every key.

## Second review pass — the migration's push was still losing the race

The migration commit above enqueued the recovered value for the server, and its own KDoc said that push is what protects it. It does not, on its own: `enqueue` is debounced 750 ms, and `withScope` runs `ensureMigrated` and then falls straight into `refreshFromServer`. On a LAN server the GET returns in ~30 ms with the registry default of 30, `applyEffectiveLocally` writes it over the value just recovered, and the sentinel is already committed so the migration never runs again. **Deterministic, not a rare race** — and it loses the value for exactly the users the migration exists to protect.

`refreshFromServer` now flushes pending writes before it reads, which is what `setSubtitleDeviceOverrideEnabled` already did ten lines away. This also closes the general case: any local write that had not yet flushed was reset by the next refresh. When the flush fails the GET almost always fails with it, and the existing early return leaves local values untouched.

The legacy-SharedPreferences import had the same hole and no push at all — it wrote recovered values into local slots that the next refresh would overwrite with registry defaults. It now enqueues what it imports, clamped to the server's accepted range so an out-of-band legacy value is stored rather than 400'd and silently dropped.

**The KDoc's stated reason for not fixing the root cause was wrong.** It claimed a `hasDeviceOverride` gate on the apply loop would break the subtitle-appearance reset. That block sits *outside* the loop and reads `hasDeviceOverride` itself, so gating the loop would not touch it. The real objection is `resetAllDeviceSettings`, which deletes every device row and refreshes precisely to pull the defaults back down — a gate would make reset a no-op. Corrected in place so the next reader isn't misled.

`resetDeviceSetting(key)` — listed below as a follow-up — is fixed here too, since it is the only path that can reach the server with a caller-supplied key.

**Test coverage.** Every existing migration test ran with `deviceId = null`, which makes `keyPrefix` empty. In production `getDeviceId` always resolves, so the scoped prefix is the only path that ships and it had zero coverage. Added. The two new ordering tests were confirmed to fail without the fix, with the exact symptom: `expected:<15> but was:<30>`.

## Correcting the issue's framing

Current server `main` already aliases the old Android spelling via `canonicalDeviceSettingKey`, with legacy-fallback reads and dual delete. So this was **not** active data loss against a current server — the rename drops Android's dependence on a deprecated compatibility shim so that shim can eventually be retired. #376's "never reaches the server" was only true against pre-alias servers.

## Verification

```
$ ./gradlew :android-shared:testDebugUnitTest --tests '*AndroidPlayerSettingsStoreTest*' --rerun-tasks
BUILD SUCCESSFUL

tests=31 failures=0 errors=0 skipped=0
```

Negative control — with the `flushNow()` removed from `refreshFromServer`, two tests fail
and nothing else does:

```
AndroidPlayerSettingsStoreTest > refreshFromServer pushes the migrated value before it reads FAILED
    java.lang.AssertionError: The refresh must flush the migrated value before reading,
    or it reads the default over it. expected:<15> but was:<30>
AndroidPlayerSettingsStoreTest > legacy cache imports are pushed so the next refresh cannot
    overwrite them FAILED
```

## Out of scope, worth follow-ups

- **A fifth drift of the same family:** the server registers `player.orientation_mode` as `{landscapeLocked, rotateFreely}`, but `orientationModeFlow` defaults to `"auto"` and `setOrientationMode` writes unvalidated — any value outside those two 400s
- ~~`resetDeviceSetting(key)` forwards an arbitrary key to `enqueueDelete` with no membership check.~~ **Fixed in the review commit** — it now requires membership in `DeviceSettings`.
- `match_frame_rate` and `sleep_timer_default_minutes` are registered in the contract manifest (Silo-Server/silo-server#479), so they return to the sync list at that cutover. Temporary churn, but they are rejected today
- The nine `subtitle.*` names had zero readers and zero writers; all Android subtitle styling already goes through the `subtitle_appearance` blob, which the server does register. Unifying them is follow-up work

Part of Silo-Server/silo-server#376. Server-side contract: Silo-Server/silo-server#479.

### AI Disclosure

- **Tool(s):** Claude Code
- **Model(s):** `claude-opus-5[1m]`
- **Involvement:** AI-assisted, under maintainer direction
- **Adversarial review:** Reviewed by a parallel multi-agent pass plus an independent adversarial pass using Codex. Both independently flagged that the migration's push could not win against the refresh it was meant to survive. The parallel pass additionally found the legacy-cache import writing without enqueuing, the unclamped value, and that every migration test ran on the unscoped key path that never ships. Each fix's test was checked to fail without the fix. One reviewer proposal was investigated and **rejected**: gating `applyEffectiveLocally`'s loop on `hasDeviceOverride` — correct that the KDoc's stated objection was bogus, wrong that the gate is safe, because it would silently break `resetAllDeviceSettings`.

Test results were re-run independently rather than taken from the implementing agent's report.

## Summary by CodeRabbit

* **Improvements**
  * Improved playback settings migration so legacy values are preserved, validated, and synchronized correctly.
  * Prevented recent setting changes from being overwritten during server refreshes.
  * Clarified which playback options remain profile-local and are not synchronized.
  * Invalid device-setting reset requests are now rejected.

* **Bug Fixes**
  * HDR10 fallback now defaults to off on Android and Android TV.
  * Playback speed and next-up prompt timing now enforce server-supported limits.
  * Match frame rate and sleep timer changes no longer trigger unnecessary server updates.

### DIFF
--- a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/settings/AndroidPlayerSettingsStore.kt
+++ b/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/settings/AndroidPlayerSettingsStore.kt
@@ -69,6 +69,11 @@ class AndroidPlayerSettingsStore(
             }
         val migrationSentinel: String =
             if (keyPrefix.isEmpty()) MIGRATION_SENTINEL_LEGACY else "migration_v2_$keyPrefix"
+
+        // Separate from [migrationSentinel] on purpose: existing installs have
+        // already recorded that one, so a rename migration folded into it would
+        // never run for exactly the users who need it.
+        val nextUpRenameSentinel: String = "migration_nextup_key_v1_$keyPrefix"
     }
 
     // Re-derive scope on every (profile or server) change.
@@ -95,6 +100,13 @@ class AndroidPlayerSettingsStore(
         }
 
     private suspend fun ensureMigrated(scope: Scope, store: DataStore<Preferences>) {
+        // Must run before the sentinel check below, and before anything else
+        // touches the store: every read path (profileScopedFlow) and every
+        // write path (withScope, and therefore refreshFromServer) funnels
+        // through here, so this is the one place that is guaranteed to happen
+        // first.
+        migrateNextUpPromptKey(scope, store)
+
         val token = scope.profileId + "/" + scope.migrationSentinel
         if (synchronized(migrationDone) { token in migrationDone }) return
         val sentinelKey = booleanPreferencesKey(scope.migrationSentinel)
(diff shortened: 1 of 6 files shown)


Quick104:
Closing this PR as superseded by #119 and the current settings implementation in #146.

The branch is no longer suitable to merge as-is because the generated contract bindings and canonical settings API changed the relevant architecture. The remaining actionable findings have been split into focused issues:

- #148 — clamp playback speed to the contract maximum
- #149 — align Dolby Vision Profile 7 fallback defaults with the contract
- #150 — reject reset requests for non-remote device keys
- #151 — prevent refresh from overwriting migrated or pending values

The old manual contract-table changes, and the changes that made match-frame-rate and sleep-timer local-only, should not be carried forward.

PR #105, opened 2026-07-26
feat(watch-together): give the feature a way in, and fix what had never run

RXWatcher:
## Why

Watch Together was fully built and **completely unreachable**. The room machinery — create/join, suggestions and voting, the WebSocket transport, `RoomSyncEngine` — compiled, and both apps routed to the lobby and the synced player, but nothing could open either.

- **Phone** was gated behind `CLIENT_WATCH_TOGETHER_SURFACE_ENABLED = false`, with the entry sheet already wired into `AppNavigation`.
- **TV had no entry point at all.** `TvItemDetailScreen` declared an `onWatchTogether` callback and `TvAppNavigation` wired it to the lobby and synced-player routes, but **no code path ever invoked it** and `TvWatchTogetherEntryDialog` was never instantiated anywhere.

Because no user could reach any of it, none of this code had ever run. Reading it turned up four defects, two of which would have broken every room in its first seconds.

## The entry points

TV gets the missing half: a **Watch Together** action in the detail overflow that opens the entry dialog, hands off to the join-code dialog, and routes on the resolved room exactly as the phone does — host with a selection goes straight to the synced player, everything else to the lobby.

Hosting targets whatever Play would launch, so a series or season hosts its **next-up episode** rather than the container, which cannot be a room selection. Audiobooks are excluded: a room's transport drives a video player and the lobby's "now playing" has nothing to show for one. Movies previously had no overflow button at all, so the action brings one with it.

`CLIENT_WATCH_TOGETHER_SURFACE_ENABLED` now gates both platforms together, so the surface can still be withdrawn in one edit.

## The four defects

**1. Frames sent before the socket was up vanished silently.** `WatchTogetherRealtimeClient.sendText` was `session?.send(...)` — no error, no queue, no retry. The interface documented the opposite ("repository gates on connection"); nothing did. Sends now report whether they reached an open session.

**2. `attach_session` could be lost for the life of the room.** Both controllers set the sent-latch *before* awaiting the send, and the only path that re-arms it is a server-had-our-session → hasn't transition, which cannot happen for a session the server never had. The latch is now set only on a delivered frame.

This one matters more than it looks. The server's readiness barrier keys entirely on the attached session — `activeParticipantCountLocked`, `allParticipantsReadyLocked` and `handleWaitingDeadline` all skip members with `sessionID == ""`. A member whose attach was dropped isn't someone the room waits for; it's someone the room doesn't know is watching. Playback starts without them and nothing logs an error.

**3. The clock stayed unsynced for up to 15s.** The ping loop fires immediately and races the handshake; on a dropped first ping no pong arrives until the next interval. `RoomSyncEngine`'s documented fallback with no offset is to apply commands *immediately*, discarding the scheduled-execute barrier that keeps members in step — during exactly the window where someone presses play. Pings now retry at 1s until the socket is up and an offset exists, then settle to the normal interval.

**4. Room-scoped calls addressed an empty room id.** They resolved `roomId ?: ""`, turning "there is no room" into a request against a malformed path whose 404/405 surfaced as though the server had rejected the user's action. They now fail fast with `no_active_room` — and fall back to the id `connect()` was called with, so voting still works in the windows where the snapshot is legitimately absent (before the first frame, and after a transient close).

## Also

- `votedIds` was an unsynchronised `MutableSet` written from an IO dispatcher and the WebSocket fold at once; it is a `StateFlow<Set<String>>` updated atomically now.
- The state-report cadence gate had been hand-copied into the TV controller with a comment asking the two to be kept in lockstep. It moves to `shared` beside `roomTransportAuthorized` — a room is one protocol, and two platforms disagreeing about when to report is a desync that only appears with two devices in a room. Its test moves to `commonTest`.
- `MAX_RECONNECT_ATTEMPTS` documented the opposite of the code. The behaviour is deliberate and load-bearing (there is a test for it); only the comment changed.

## Testing

`assembleDebug` and the full unit suites green on `shared`, `androidApp` and `androidTvApp`, from a clean `--rerun-tasks` run on this branch based on `main`.

New tests:
- `TvWatchTogetherSurfaceSourceTest` — the TV failure mode was *absence* of wiring, which nothing behavioural could catch, because the dialog was not reachable to observe.
- Repository tests asserting sends report undelivered before the socket is up, and delivered once connected.
- `RoomStateReportGateTest` moves to `commonTest` with the gate.

**Not yet exercised against a live server.** The TV build is installed on an emulator, but a room needs two participants and I have not run one end to end. The server side was read closely — transport authority mirrors the client's shared `roomTransportAuthorized`, the waiting deadline is epoch-guarded, persist conflicts restore and re-arm rather than dropping the transition — and no server changes were needed.

Worth flagging for whoever picks this up: there is **no Apple counterpart**. `silo-apple` has no Watch Together at all, so this surface has no gold-standard client to mirror and anything ambiguous about its shape is a product decision rather than a parity bug.

## AI-use disclosure

Written by Claude Opus 5 (Claude Code) working with @RXWatcher, who directed the work and reviewed it. The defects were found by reading the code, not by running it.

## Summary by CodeRabbit

* **New Features**
  * Enabled Watch Together from Android TV item detail menus, with host/join flows and lobby navigation.
* **Bug Fixes**
  * Improved delivery handling so playback updates are not treated as sent when disconnected.
  * Added faster clock-sync retries until synchronization is established.
  * Prevented room requests from being sent without an active room.
  * Improved preservation of voting state across updates and reconnects.
* **Tests**
  * Expanded coverage for Watch Together visibility, navigation, delivery failures, and room-state reporting.


RXWatcher:
Closing — opened against the wrong remote. This work lives on the RXWatcher fork.

PR #106, opened 2026-07-26
feat(watch-together): give the feature a way in, and fix what had never run

RXWatcher:
## Why

Watch Together was fully built and **completely unreachable**. Create/join, suggestions and voting, the WebSocket transport, `RoomSyncEngine` — all compiled, and both apps routed to the lobby and the synced player, but nothing could open either.

- **Phone** was gated behind `CLIENT_WATCH_TOGETHER_SURFACE_ENABLED = false`, with the entry sheet already wired into `AppNavigation`.
- **TV had no entry point at all.** `TvItemDetailScreen` declared an `onWatchTogether` callback and `TvAppNavigation` wired it to the lobby and synced-player routes, but **no code path ever invoked it** and `TvWatchTogetherEntryDialog` was never instantiated anywhere.

Because nobody could reach it, none of this code had ever run. Opening the door turned up six defects, four of which broke every room.

## The entry points

TV gets the missing half: a **Watch Together** action in the detail overflow that opens the entry dialog, hands off to the join-code dialog, and routes on the resolved room as the phone does. Hosting targets whatever Play would launch, so a series or season hosts its **next-up episode** rather than the container, which cannot be a room selection. Audiobooks are excluded — a room's transport drives a video player. Movies had no overflow button at all, so the action brings one with it.

## What was broken underneath

**1. Frames sent before the socket was up vanished silently.** `sendText` was `session?.send(...)` — no error, no queue, no retry. The interface documented the opposite ("repository gates on connection"); nothing did. Sends now report whether they reached an open session.

**2. `attach_session` could be lost for the life of the room.** Both controllers set the sent-latch *before* awaiting the send, and the only path that re-arms it is a server-had-our-session → hasn't transition — impossible for a session the server never had.

This is worse than it looks. The server's readiness barrier keys entirely on the attached session: `activeParticipantCountLocked`, `allParticipantsReadyLocked` and `handleWaitingDeadline` all skip members with `sessionID == ""`. A member whose attach was dropped isn't someone the room waits for, it's someone the room doesn't know is watching — playback starts without them and nothing logs an error.

**3. The clock stayed unsynced for up to 15s.** The ping loop fires immediately and races the handshake; on a dropped first ping no pong arrives until the next interval. `RoomSyncEngine`'s documented fallback with no offset is to apply commands *immediately*, discarding the scheduled-execute barrier that keeps members in step — during exactly the window where someone presses play. Pings retry at 1s until the socket is up and an offset exists.

**4. Room-scoped calls addressed an empty room id**, turning "there is no room" into a request against a malformed path whose 404/405 surfaced as a rejected action. They fail fast now, and fall back to the id `connect()` was called with, so voting still works before the first snapshot arrives and after a transient close.

**5. The invite code was unreadable.** Hosting pre-selects the title, so a new room is playing-and-selected the instant it exists — and every routing decision keyed on that, so creating a room flashed the code for a fraction of a second before the player replaced it. A host who is still alone now stays in the lobby; it hands off once the member count rises. Guests are unaffected.

**6. The lobby could not reach its own controls.** A single non-scrolling column, taller than a 1080p panel, so "Guest controls", "Close room" and the suggestion list were drawn past the bottom edge. It is now two columns: the invite (code + QR) pinned left and never scrolling, everything focusable in a right column that scrolls — the code has to stay on screen while you read it out.

**Also:** the QR had no quiet zone (`MARGIN` was 0, and the promised padding did not exist — the canvas matched the box), and a rounded clip cut the corners off the finder patterns. Squared, with the spec's clear border. `votedIds` was an unsynchronised `MutableSet` written from an IO dispatcher and the WebSocket fold at once. The state-report cadence gate had been hand-copied into the TV controller with a comment asking the copies be kept in lockstep — it moves to `shared` beside `roomTransportAuthorized`.

## Testing

`assembleDebug` and the full unit suites green on `shared`, `androidApp` and `androidTvApp`, from this branch based on `upstream/main`.

New tests: `TvWatchTogetherSurfaceSourceTest` (the TV failure was *absence* of wiring, which nothing behavioural could catch — the dialog was not reachable to observe); repository tests that sends report undelivered before the socket is up; lobby hand-off tests for host-alone / host-with-guest / guest, verified to fail when the guard is removed.

**Not exercised end to end against a live server.** The TV build is installed on hardware and the lobby, invite code and QR were checked visually, but a room needs two participants and I have not run one. The server side was read closely — transport authority mirrors the client's shared `roomTransportAuthorized`, the waiting deadline is epoch-guarded, persist conflicts restore and re-arm — and no server changes were needed.

Worth flagging: there is **no Apple counterpart**. `silo-apple` has no Watch Together at all, so this surface has no gold-standard client to mirror, and anything ambiguous about its shape is a product decision rather than a parity bug.

## AI-use disclosure

Written by Claude Opus 5 (Claude Code) working with @RXWatcher, who directed the work, reviewed it, and did the on-device checks. The defects were found by reading the code, not by running it.

## Summary by CodeRabbit

* **New Features**
  * Watch Together is now available from supported detail-screen menus on TV.
  * Create vote rooms, submit suggestions, identify winning titles, and start the winner.
  * TV users can suggest titles to an active room.
  * Improved room navigation keeps solo hosts in the lobby until others join.
* **Bug Fixes**
  * Improved room connection reliability and playback synchronization.
  * QR codes now include clearer spacing for easier scanning.
  * Audiobook items do not display unsupported Watch Together actions.


RXWatcher:
Pushed the voting client work onto this branch, paired with Silo-Server/silo-server#486:

- Hosting now offers **Host a vote room** alongside the ordinary one. `selection_mode` is fixed at creation server-side, so it has to be chosen there rather than switched on later.
- A vote room opens **empty**. Both clients previously created the room and immediately set the title the host was looking at as its selection, which put the room straight into `Playing` and threw everyone into the player before a single suggestion existed — the vote never happened and the mode was dead on arrival. `host_pick` still pre-selects, because there the host's choice *is* the selection.
- The lobby marks the leading suggestion **WINNING** and gives the host one `Start winner: <title>` action above the list, disabled until something has a vote (the server refuses that start, and a button that can only fail is worse than one visibly not ready).
- The per-row **Pick** is hidden in vote rooms, since the server promotes only the winner.
- `roomVoteWinner` lives in `shared` and mirrors the server rule exactly — head of the vote-ordered list, nothing when no votes are cast — so the client's idea of the leader cannot drift from the server's and leave a host staring at a 409 they cannot explain.

Full unit suites and `assembleDebug` green on this branch. Still not exercised with two participants in a real room.

RXWatcher:
Pushed three more commits, which together make a room something you stay *in* rather than a screen you are *on*.

**Suggesting.** `addSuggestion` existed on the repository and the API and was called from nowhere on either platform, so a vote room could never acquire candidates — the lobby sat on "No suggestions yet" permanently and Start winner could never leave its disabled state. The TV detail overflow now offers **Suggest to Watch Together** whenever you are actually in a room. It belongs on item detail rather than in the lobby: you choose something to suggest by browsing for it, and the lobby has no catalogue in it. *Phone has no equivalent yet.*

**One owner for the connection.** All four screen-scoped connections are gone — the lobby view models and both `RoomSyncController`s now adopt an app-scoped `RoomSession` instead of opening their own. This fixes two things:

- A room survives leaving the lobby. The socket used to live in `viewModelScope`, so backing out closed it, the server saw the host disconnect, and the room closed for everyone. Browsing to suggest something took the room down with it.
- A lobby → player hand-off no longer runs **two** reconnect loops. Both scopes called `connect()`, racing over the repository's single `realtime` field; whichever lost left a dead client that believed it was connected. That was a live bug independent of this feature — one owner removes it by construction rather than by timing.

Leaving is explicit now: only the lobby's `leave()` drops the connection, never composition ending.

Paired with Silo-Server/silo-server#487, which raises the server's host-disconnect grace from 15s to two minutes. They cover different failure modes — this stops us reporting "host left" when the host merely navigated; that one covers drops the client cannot prevent (backgrounding, connectivity, process death).

Unit suites and `assembleDebug` green. **None of these three commits has been run against a live room.** The hoist in particular changes who owns the socket while `RoomSyncController` still assumes it drives attach and session lifecycle — the seam between those two is where I would look first.

RXWatcher:
Added the mobile half of suggesting, so the phone is no longer a second-class member of a vote room — it could vote but never nominate.

**Suggest to Watch Together** now appears in the phone's movie *and* series detail overflow whenever the viewer is in a room, with the same toast confirmation as TV. Two composables needed it — `MovieDetailContent` and `SeriesDetailContent` have separate overflow menus, and doing only the first would have compiled while silently leaving series detail without the action.

This only works because of the session hoist earlier in this branch: the room is read from the repository's live snapshot, which now outlives any single screen. Before that, the action could only have appeared while the lobby was composed — i.e. never, since you are on a detail page when you want it.

Full unit suites plus both `assembleDebug` targets green on this branch, based on `upstream/main`.

One cherry-pick conflict resolving onto this branch, worth naming since it is the kind of thing that hides a mistake: an import block where `main` had picked up an unrelated `DownloadRecord` import this branch predates. Resolved to keep only the import this commit needs; the build confirms nothing else wanted it.

RXWatcher:
Closing as superseded by the reviewed Watch Together work that has since landed through the forward-split stack and subsequent profile-entry changes. Retaining this PR and its branch as historical reference; it is not being merged.

PR #108, opened 2026-07-26
Integrate the fork: subtitles/PGS, Watch Together, security hardening, player and sync fixes

RXWatcher:
## What this is

Everything that has accumulated on my fork's `main` and never been proposed upstream — 80 commits, 320 files. It is merged with current `upstream/main`, builds clean, and the full unit suites pass on `shared`, `androidApp` and `androidTvApp`.

I would rather have sent this as a dozen focused PRs, and I tried. It does not decompose: the history contains an earlier upstream integration merge, several cross-cutting refactors (the player coordinator, the libass handler lifecycle, the subtitle transaction model) that later commits build directly on, and a shared-model rename resolved in this merge. Cherry-picking themes out of it produces branches that do not compile. Splitting it honestly would mean rewriting the work, not reorganising it.

So: take it as an integration branch, or tell me which slices you want carved out and I will rebuild those properly rather than pretend a `git cherry-pick` does it.

## What is in it, by theme

**Subtitles and PGS (~14 commits).** A `.sup` extractor so bitmap sidecars can be framed at all (Media3 parses PGS but only one display set at a time, timestamped by a container that a raw `.sup` does not have). Client-side PGS mounting instead of forcing server burn-in. SDH picks matching their own artifact. Per-item subtitle sync, a burn-in guard, a selectable CC row, in-stream captions named rather than numbered. Appearance defaults aligned to the web client. Letterbox insets so cues sit over the picture on sources with baked-in bars — this is the client half of Silo-Server/silo-server#483.

**Watch Together (7).** The feature was fully built and completely unreachable: no entry point existed on TV at all, and the phone's was behind a const. Opening it exposed six defects, including client→server frames vanishing silently when the socket was not yet up, which cost the room `attach_session` for its whole lifetime. Broken out for focused review as #106.

**TV request detail (1).** Content rendered under the shell's nav overlay. Broken out as #107.

**Security and supply chain (~15).** Consent before cleartext login; rejecting a scope whose persistent credentials were replaced; identity-scoped download paths and the races around them; ambiguous HTTP authority rejection; keeping credentials on the Silo origin; authenticated PiP actions; epub sanitisation with a parsed allowlist, SVG paint validation, bounded remote/archive content, webview isolation; dependency verification metadata, pinned native provenance, fail-closed OSV parsing.

**Player and sync correctness (~20).** Sessions that finish starting after the user leaves; unsettled publications wedging future starts; failed final-position writes; libass handlers accumulating embedded fonts; outbox ordering; draining on profile change; purging removed servers.

**TV player and picker fixes (~10), plus test infrastructure (~5)** — including a recorded emulator sweep.

## The merge

Non-trivial, and worth a look:

- Upstream's #102 deleted the source-grep pseudo-tests. I accepted every one of those deletions (16 files) rather than reinstating them.
- Upstream renamed `PlaybackSourceV3` → `PlaybackSourceDescriptorV3` and added `durationSeconds`; my fork had added `colorRange` and the letterbox fractions to the old type. The auto-merge produced a data class with **two** `source` fields. Resolved to a single type on upstream's name carrying all the fields.
- `PlaybackApiTest`: upstream replaced a hand-maintained feature list with `PLAYBACK_START_CLIENT_FEATURES_V3`. Took upstream's.

## Caveats, stated plainly

- **Watch Together has never been run with two participants.** Everything else here has been exercised on real hardware (NVIDIA Shield) against a live server over an extended period.
- `TvWatchTogetherSurfaceSourceTest` in the Watch Together work is a source-grep test — the pattern #102 just removed. It exists because the failure mode was *absence* of wiring, which nothing behavioural could observe while the dialog was unreachable. Happy to drop it if the policy is absolute.
- Two overlapping PRs (#106, #107) contain subsets of this branch, kept open because they are small and self-contained. Close them if you prefer to review this whole.

## AI-use disclosure

Written by Claude Opus 5 (Claude Code) working with @RXWatcher, who directed the work, reviewed it, and ran the device testing throughout.


RXWatcher:
Refreshed to current `main`. Adds, since the last push: the vote-room client work, the suggest action, and the app-scoped room session that gives the room a single owner above the screens.

The Watch Together portion of this branch is reviewable on its own as #106, which now carries the same commits and a fuller write-up. Server counterparts: Silo-Server/silo-server#486 (vote enforcement) and #487 (host-disconnect grace).

RXWatcher:
Refreshed to current `main`. Since the last push this adds the Watch Together host override plus three Apple-parity ports:

- **libass subtitles inside the title-safe area** — ports Silo-Server/silo-apple#95. Media3's text layer clears the bottom via `setBottomPaddingFraction`, but libass places ASS/SSA cues against whatever surface it is handed, so a cue authored at the bottom of the frame was composited into pixels the panel never shows. The line simply never appeared — no error, no log. The inset goes on the surface so both renderers inherit it, applied after the letterbox inset so the two compose (bars first, then overscan), and the position presets give back what the surface took so styled text does not move.
- **manual Up Next control** (also standalone as #110)
- **mixed library visibility** (also standalone as #111)

The title-safe fix has no standalone PR because it builds on the letterbox/PGS subtitle work, which only exists on this branch — it does not apply to `upstream/main` alone. If the subtitle stack is split out for review, that fix should ride with it rather than travel separately.

RXWatcher:
Closing without merge: this integration PR has been superseded by the reviewable forward-split stack, beginning with draft PRs #112 and #113. PR #108 is retained as the archival integration reference for traceability.

PR #147, opened 2026-07-30
fix(playback): bound the load-idle window so proxies stop dropping stream connections

RXWatcher:
## The bug

Long direct-play sessions lose their connection every few minutes. The buffer fills, the client stops reading, the proxy's client-facing `send_timeout` closes the socket, and the next read finds it dead.

`DefaultLoadControl` fills to `maxBufferMs`, then stops reading the socket until the buffer drains below `minBufferMs` — so **`max - min` is literally how long the TCP connection sits idle**. The shipped policy was min 50s / max 120s: a **70-second idle window** against a 60-second `send_timeout`. That is arithmetic, not a race.

It is independently corroborated by a comment in the openresty config:

> `send_timeout` … *was 3600s for a while because 60s killed idle-but-alive progressive stream sockets under player backpressure (**reconnect hiccup every ~2-3 min**)*

A 120s buffer fills in about two to three minutes. The observed symptom and the arithmetic agree.

## The fix

`maxBufferMs` is no longer written by hand:

```
maxBufferMs = minBufferMs + MAX_LOAD_IDLE_MS
```

enforced by `init { require(...) }`, so neither the public constructor nor `copy()` can reintroduce a wide window. Depth and idle window become independent: the buffer can hold 180s while the socket is never quiet for more than the window.

**`MAX_LOAD_IDLE_MS = 15_000`, not 30_000.** The invariant is expressed in *media* time but a proxy measures *wall clock*, and `DefaultLoadControl` scales `minBufferUs` only for speeds **above** 1.0. Audiobooks offer 0.5x and share this load control, so a 30s media window would become 60s of wall clock — exactly the timeout. 15s is that 30s wall-clock budget scaled down by the slowest selectable rate (`30_000 * 0.5`); it converts back by dividing, `15_000 / 0.5 = 30_000` of wall clock.

## Memory budget: half the heap

Measured via adb, not assumed:

| Device | `memoryClass` | before | after |
|---|---|---|---|
| Google TV Streamer | 384 MB | 160 MiB | 192 MiB |
| NVIDIA Shield | 192 MB | 96 MiB | 96 MiB |
| low-end | 96 MB | 48 MiB | 48 MiB |

A 25% rule was tried first and *reduced* the budget on every real device — the opposite of the intent. Low-RAM devices take the smaller of 24 MiB and their proportional share, so a small heap is never handed half of itself.

## Also fixed

- The byte clamp no longer silently overrides the stated depth. Previously a 60 Mbps stream on a low-RAM device reported 20s while holding ~6s; the budget is now authoritative and the reported depth is the truth.
- Depth/bytes transposition is a **compile error** (`@JvmInline value class BufferDepthMs` / `BufferTargetBytes`) — the Media3 override can't be reached by tests without `Parameters` scaffolding, so the mistake is made unrepresentable instead.
- `computeBufferSizing` is a pure function; the Media3 override is a thin adapter with no policy arithmetic in it.
- The dead `PlaybackBufferMode` enum and its wire parsing are removed — `SiloPlayerFactory` hardcoded `Balanced`, so two of the three modes were unreachable.

No transcode special case: the server's `TranscodeThrottler` already bounds transcode-ahead, and a deep target on a realtime-bound encoder is inert rather than harmful.

## Verification

Release build (R8 + resource shrinking) installed on a Shield and a Google TV Streamer.

- **No R8 breakage** — the risk this repo flags for minified release builds.
- **Both buffering regimes observed live**: high-bitrate content byte-bound at ~172 MiB heap; low-bitrate time-bound at ~62 MiB reaching the full 180s depth.
- **Against a live 60s `send_timeout`** (the 3600s workaround was reverted on 2026-07-28), a direct-play session produced a **70.1-second request that completed with `206`**, plus ~9 minutes of direct play with no reconnects — where the old build hiccupped every 2-3 minutes. `send_timeout` applies *between successive writes*, so a long request completing means no single idle gap approached 60s.

Unit tests: 1,005 passing; both `:androidApp:assembleDebug` and `:androidTvApp:assembleDebug` build.

## Review round (CodeRabbit + the two follow-ups this body originally listed)

- **The memory budget is now authoritative when it falls below the byte floor.** `calculateBitrateTargetBufferBytes` requires `maximumBytes >= minimumBytes`; that was being restored by raising the ceiling to meet the floor, so a device budgeted under 16 MiB could be handed a byte target larger than the heap it was allowed. It now lowers the floor instead. Unreachable on shipping hardware — both constants are exactly 16 MiB today — which is why it got a regression test rather than a comment.
- **The idle-window guard is strict.** It asserted `stretchedWallClockMs <= ASSUMED_PROXY_SEND_TIMEOUT_MS`, so reverting `MAX_LOAD_IDLE_MS` to the pre-fix 30_000 gave `60_000 <= 60_000` and passed — the one test whose job was to catch that revert could not fail. Verified by setting the constant back to 30_000 and watching it fail.
- **Doc drift corrected.** The spec and plan still said `MAX_LOAD_IDLE_MS = 30_000` and described the 20s depth floor as a guarantee. The constant became 15_000, and the floor is deliberately *not* a guarantee: a known bitrate whose budget funds less than 20s yields the honest smaller number, which is the point of making the reduction explicit.

## Still open (non-blocking)

- An A/B against the pre-fix build on the same title would make the field evidence airtight; the current evidence is a single session plus the config's own account of the old behaviour.

Design and plan are in `docs/superpowers/` on this branch.

### DIFF
--- a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackBufferPolicy.kt
+++ b/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackBufferPolicy.kt
@@ -1,17 +1,10 @@
 package org.siloserver.silo.common.player
 
-enum class PlaybackBufferMode(val wireValue: String, val label: String) {
-    QuickStart("quick_start", "Quick start"),
-    Balanced("balanced", "Balanced"),
-    SmoothPlayback("smooth_playback", "Smooth playback");
-
-    companion object {
-        fun fromWire(value: String?): PlaybackBufferMode = entries.firstOrNull {
-            it.wireValue == value
-        } ?: SmoothPlayback
-    }
-}
-
+/**
+ * Buffering policy derived from what the player can observe. There is no user
+ * setting and no server-supplied mode: the numbers follow from the device and
+ * the stream.
+ */
 data class PlaybackBufferPolicy(
     val minBufferMs: Int,
     val maxBufferMs: Int,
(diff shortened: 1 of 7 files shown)


RXWatcher:
@coderabbitai review

PR #152, opened 2026-07-30
feat(settings): offer every language the web does, and make them match

RXWatcher:
## The report

> On Android I'm running the store version but can't put Dutch as default sub.

Correct, and it is not fixed on `main` either — the picker offers the same ten languages in both, and Dutch is in neither:

`English, Spanish, French, German, Japanese, Korean, Chinese, Portuguese, Italian, Russian`

## It was never a server limitation

The settings contract types `playback.subtitle_language` as:

```json
{ "type": "language_tag", "nullable": true }
```

validated for BCP 47 shape only — there is no enum. The web client already offers **37** languages, Dutch included. Android's picker was simply shorter.

`LanguageOptions.tags` now mirrors the web's ISO 639-1 table (`web/src/player/utils/languageNames.ts` in silo-server) in its order. Ten becomes 37.

Worth noting: the table's own comment already named `"nl"` as a tag that arrives *"synced from another surface"* and must be preserved. So a Dutch preference set on the web was being honoured by playback the whole time — the picker just refused to display or re-select it. The user was looking at a setting that was working and appeared absent.

## Completing the picker exposed an older bug

Auto-selection compares the stored preference against a track's language with **both sides** through `canonicalSubtitleLanguage` (`MobileSubtitleAutoSelection`, `TrackSelectionFingerprint`). That function's ISO 639-2 alias table covered seven languages while the picker offered ten:

```kotlin
"eng" -> "en"   "spa" -> "es"   "fre", "fra" -> "fr"
"ger", "deu" -> "de"   "dut", "nld" -> "nl"   "jpn" -> "ja"   "dan" -> "da"
```

A track tagged `ita` canonicalises to `ita`, which never equals a stored `it`. **Italian, Portuguese, Korean, Chinese and Russian were already offered in the picker but could not auto-select on any server exposing bibliographic codes.** Shipping 27 more languages against that table would have multiplied the problem.

The alias table now covers all 37, carrying both bibliographic and terminological forms where ISO 639-2 has both (`cze`/`ces`, `gre`/`ell`, `may`/`msa`, `per`/`fas`, …), plus the pre-1989 codes `iw` → `he` and `in` → `id` that older muxing tools still emit. Norwegian folds `nor`/`nob`/`nno` into the single offered `no`, since Bokmål and Nynorsk are one choice in the picker.

Dutch itself was already mapped (`dut`, `nld` → `nl`), so the reported case needed only the picker entry — but it would have been the only one of the 27 additions that worked.

## Tests

`SubtitleLanguageTest` (new, 9 cases) and `LanguageOptionsTest` (10) both pass, along with the full `:shared` suite; `:androidApp:assembleDebug` and `:androidTvApp:assembleDebug` both build.

The two that would catch a regression here:

- **every picker tag is already canonical** — an option that canonicalises to something else can never equal a matching track
- **every offered language is reachable from its 639-2 form** — if a language is in the picker, some bibliographic spelling of it must canonicalise back to the offered tag, or a server exposing that spelling can never match the choice

Plus the reported case directly: Dutch matches across `nl`, `nld`, `dut`, `NL`, `nl-NL`, `nl_BE`.

## Parity note

**silo-apple offers 12** — these ten plus Arabic and Hindi — and also cannot select Dutch. It needs the same treatment; this PR does not touch it.

---

### The alias table now mirrors the server instead of paraphrasing it

The first version of the map was hand-rolled and had already diverged on its third entry. The server's `canonical_language_code` (migration 121, mirroring `internal/lang`) keeps `nob` → `nb` and `nno` → `nn` **distinct**; the hand-rolled version folded both into `no`. The catalog stores what the server canonicalised, so that is not cosmetic — a Bokmål file is `nb` server-side and would have been compared against a client-side `no`, matching nothing.

It is now a faithful copy of all 90 server mappings. That covers far more languages than the pickers offer, which costs nothing and means an unlisted language synced from another surface still compares correctly.

Kept deliberately beyond the server's table: `iw` → `he` and `in` → `id`, the pre-1989 639-1 spellings Java's `Locale` still produces. These are 639-1 rather than 639-2, so the server's map does not carry them and passes them through unchanged — **worth adding server-side**, flagged in a comment.

### Language filter facets show names, not raw codes

`CatalogFilterState` and `TvLibraryBrowseFacets` mapped each vocabulary entry to itself, so library filters rendered `eng`, `nld`, `dut` instead of English and Dutch. New `LanguageOptions.displayLanguage` canonicalises before labelling, so all three Dutch spellings read "Dutch". A code no picker offers is returned unchanged rather than blanked — display and wire value stay the same string, and the filter still applies.

### The viewer's own languages come first

Thirty-seven entries is a long way to travel on a D-pad. The obvious shortcut — device locale — was **measured and rejected**: the Shield here ships `persist.sys.locale` empty, `system_locales` null, and `ro.product.locale` at the factory `en-US`, so locale-first would suggest English to a Dutch viewer.

The catalog is a real signal. Settings fetches `catalog/filters?include_technical=true` and lifts the languages the library actually carries to the top of all six language pickers (audio, subtitle, metadata × phone, TV). The current value is pinned above them so it stays reachable when the fetch fails. The tail is alphabetical — the web's prominence order stops meaning anything past the first ten.

The fetch is best-effort and silent on failure; the pickers work fine without it.

**Not done:** visual group headers ("In your library" / "All languages"). Both pickers render flat lists, so headers need a non-selectable row type on two platforms — worth its own change.

## Summary by CodeRabbit

* **New Features**
  * Language pickers now prioritize the current selection and languages found in your library.
  * Language codes are displayed using recognizable language names, including support for regional and script-specific codes.
  * Audio, subtitle, metadata, and catalog language filters now use consistent language labels.
  * Expanded language coverage and improved handling of alternate language codes.

* **Bug Fixes**
  * Preserved language codes that are not available in the picker instead of hiding them.
  * Prevented duplicate language options and ensured selections remain reversible.

### DIFF
--- a/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/di/AndroidModule.kt
+++ b/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/di/AndroidModule.kt
@@ -407,7 +407,7 @@ val androidModule = module {
             tmdbId = args.second,
         )
     }
-    viewModel { SettingsViewModel(get(), get(), get(), get(), get(), get(), get()) }
+    viewModel { SettingsViewModel(get(), get(), get(), get(), get(), get(), get(), get()) }
     viewModel { DiagnosticsViewModel(get()) }
     viewModel { AdminEntryViewModel(get(), get()) }
     viewModel { AdminStatsViewModel(get()) }
--- a/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/settings/PlaybackSettings.kt
+++ b/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/settings/PlaybackSettings.kt
@@ -26,8 +26,6 @@ import org.siloserver.silo.model.settings.QualityPresets
 // Audio language stores BCP 47 tags ("" = no preference) — display labels,
 // persist codes, as the server's settings contract requires. Shared with the TV
 // UI and with subtitles so the four surfaces cannot drift apart again.
-private val audioLanguageOptions = LanguageOptions.options(unsetLabel = "Default")
-private val audioLanguageLabels = audioLanguageOptions.map { it.second }
 
 // Discrete choices for the two behavior settings (0 = off). Dropdown idiom
 // matches the rest of this section; the label↔value maps below convert.
@@ -53,6 +51,8 @@ fun PlaybackSettings(
     qualityResolution: String,
     maxBitrateKbps: Int?,
     audioLanguage: String,
+    /** Codes the catalog holds; floats the viewer's own languages to the top. */
+    libraryLanguages: List<String> = emptyList(),
     autoSkipIntro: Boolean,
     autoSkipCredits: Boolean,
     pictureInPictureEnabled: Boolean,
@@ -77,6 +77,14 @@ fun PlaybackSettings(
     onResetPlaybackOverrides: () -> Unit,
     modifier: Modifier = Modifier,
 ) {
+    val audioLanguageLabels = remember(libraryLanguages, audioLanguage) {
+        LanguageOptions.optionsPrioritising(
+            unsetLabel = "Default",
+            libraryLanguages = libraryLanguages,
+            current = audioLanguage,
+        ).map { it.second }
+    }
+
     SettingsSectionCard(modifier = modifier) {
         SettingsSectionHeader("Playback")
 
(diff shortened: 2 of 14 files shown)


RXWatcher:
@coderabbitai review

RXWatcher:
Closing as superseded by the contract-driven language catalog architecture merged in #154. The remaining useful behavior—human-readable language labels in library facets—will be extracted as a focused current-main change rather than merging the conflicting hand-maintained catalog. The branch remains available for traceability.

PR #164, opened 2026-08-03
fix(tv): restore Shield focus navigation

RXWatcher:
## Summary

- preserve and restore the exact For You recommendation card after returning from detail, including reordered and horizontally offscreen targets
- retain PR #162 late first-row anchor recovery and make only the For You recommendations view use the solid top treatment already shown by Watchlist/Favorites
- route Calendar focus from shelves to weekdays to filters to the Calendar tab, including empty/error states
- make Diagnostics crash-report controls reliably focusable and prevent held D-pad repeats from skipping layers

Watchlist remains inside For You. Its behavior, navigation, layout, rendering, and generic detail callback are unchanged.

This PR includes and supersedes #162.

## Verification

- shared recommendation identity test: passed
- focused For You resolver, lifecycle, offscreen-row, and top-anchor tests: passed
- focused Diagnostics tests: passed
- full Android TV unit suite: 794 tests, 0 failures
- androidTvApp debug APK assembly: passed
- final whole-branch review and scoped fix re-review: ready to merge
- TvPersonalScreens.kt protected-file diff: empty

## Device follow-up

The APK was built but was not installed or launched during this branch pass. Shield Pro focus and visual smoke checks remain the post-PR release gate.

## Summary by CodeRabbit

* **Improvements**
  * Restores focus to previously selected recommendations after returning from details, including reordered or temporarily unavailable content.
  * Improves D-pad navigation across Calendar filters, week controls, menus, and content.
  * Enhances Diagnostics settings focus order, retries, and directional navigation.
  * Preserves the For You list’s top position after delayed focus changes.
  * Displays a solid top-bar background when needed for For You navigation.

* **Tests**
  * Added coverage for focus restoration, navigation boundaries, delayed scrolling, and diagnostics controls.

### DIFF
--- a/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/components/TvMediaRow.kt
+++ b/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/components/TvMediaRow.kt
@@ -8,6 +8,7 @@ import androidx.compose.foundation.lazy.LazyRow
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
@@ -22,6 +23,7 @@ import androidx.compose.ui.input.key.KeyEventType
 import androidx.compose.ui.input.key.key
 import androidx.compose.ui.input.key.onPreviewKeyEvent
 import androidx.compose.ui.input.key.type
+import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.ExperimentalComposeUiApi
 import org.siloserver.silo.model.section.SectionItem
@@ -98,6 +100,12 @@ fun TvMediaRow(
      *  of card 0. Callers should only pass it while a restore is pending. */
     restoreFocusIndex: Int = -1,
     restoreFocusRequester: FocusRequester? = null,
+    /** Nonzero only while an exact-card return is pending. The row uses its
+     *  private horizontal state to compose [restoreFocusIndex] before focus is
+     *  requested; ordinary row rendering never changes horizontal position. */
+    restoreFocusRequest: Int = 0,
+    onRestoreFocusTargetPlaced: ((Int, Int) -> Unit)? = null,
+    onRestoreFocusTargetDisposed: ((Int, Int) -> Unit)? = null,
     /** Fired (on focus GAIN only) with whichever card the user focuses, so the
      *  Skyline marquee + backdrop can preview the focused item. */
     onItemFocused: ((SectionItem) -> Unit)? = null,
@@ -124,6 +132,16 @@ fun TvMediaRow(
                 contentType = "${cardLayout.name}:${style.name}:${item.type}",
             )
         }
+    }
+    val restoreFocusContentId = rowItems.getOrNull(restoreFocusIndex)?.item?.contentId
+
+    LaunchedEffect(restoreFocusRequest, restoreFocusIndex, restoreFocusContentId) {
+        prepareTvMediaRowFocusRestore(
+            requestId = restoreFocusRequest,
+            restoreFocusIndex = restoreFocusIndex,
+            itemCount = rowItems.size,
+            scrollToItem = rowState::scrollToItem,
+        )
     }
 
     LaunchedEffect(firstItemFocusRequest) {
@@ -189,6 +207,15 @@ fun TvMediaRow(
                 contentType = { _, rowItem -> rowItem.contentType },
             ) { index, rowItem ->
                 val item = rowItem.item
+                val isRestoreFocusTarget =
+                    restoreFocusRequest > 0 && index == restoreFocusIndex
+                if (isRestoreFocusTarget && onRestoreFocusTargetDisposed != null) {
+                    DisposableEffect(restoreFocusRequest, index) {
+                        onDispose {
+                            onRestoreFocusTargetDisposed(restoreFocusRequest, index)
+                        }
+                    }
+                }
                 // Always anchor firstItemFocusRequester to index 0 so it can
                 // serve as a stable fallback target for focusRestorer and for
                 // imperative requestFocus() calls from parent screens.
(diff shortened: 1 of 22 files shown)


RXWatcher:
### Code review

Found 5 issues (1 blocking, 4 lower confidence):

**1. Calendar held-Up freezes mid-list — the `isRepeat` branch lost its guard (blocking)**

`isRepeat -> StayInContent` is now an unconditional catch-all. On `main` it was scoped as `focusedShelfIndex == null && isRepeat` by [`d6f2c15c`](https://github.com/Silo-Server/silo-android/commit/d6f2c15c) ("fix(tv): keep repeated up in content"), specifically so a held Up deeper in the shelf list still fell through to `MoveWithinContent`.

The branch above it already handles the boundary case inline (`if (isRepeat) StayInContent else ReturnToControls`), so the new branch contributes nothing to the case it was written for — its only effect is to swallow repeats when `focusedShelfIndex` is non-null. Holding Up on a mid-list day shelf moves once, then freezes until the button is released.

This also breaks the PR's own stated Global Constraint in `docs/superpowers/plans/2026-08-03-shield-focus-restoration.md`: "Preserve the existing one-layer-per-press behavior for repeated D-pad events."

`TvCalendarFocusRoutingTest.kt` has no case combining `isRepeat = true` with a non-null `focusedShelfIndex`, which is why this passed.

https://github.com/Silo-Server/silo-android/blob/4d30facafbb7ac9a4bf12c6a5f3eb5f811f62b02/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/screens/calendar/TvCalendarScreen.kt#L724-L734

**2. Diagnostics Up is a one-way door**

Up clamps (`order[(index - 1).coerceAtLeast(0)]`) while Down falls through (`order.getOrNull(index + 1)`). Since `consume = target != null`, clamping always consumes, so every Up press on the first crash-report option is swallowed. There is a STATUS section above it in the `LazyColumn`, so there is a legitimate focus target being blocked. The test name `firstChoiceHoldsAtUpperBoundary` suggests this may be deliberate — flagging because it is asymmetric with Down.

https://github.com/Silo-Server/silo-android/blob/4d30facafbb7ac9a4bf12c6a5f3eb5f811f62b02/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/screens/settings/diagnostics/TvDiagnosticsSettingsScreen.kt#L266-L283

**3. `stableSectionId()` fallback drops the uniqueness guarantee**

The old `discover_${index}_${type}` was unique by construction. The fallback used when the server does not send `sectionKind` is `discover:type=$type:label=$label`, which is not. There is no `distinctBy` in `toResolvedSections()`, and `section.id` is the `LazyColumn` key on both TV (`TvRecommendationsScreen.kt`) and phone (`RecommendationsScreen.kt`) — duplicate keys throw in Compose.

Requires an older server emitting two rows sharing both type and label. `RecommendationsSectionIdentityTest` only covers the `sectionKind`-present path. Neither design doc mentions the coordinated server field or the degrade path. (No upgrade risk — nothing persisted is keyed on the old format.)

https://github.com/Silo-Server/silo-android/blob/4d30facafbb7ac9a4bf12c6a5f3eb5f811f62b02/shared/src/commonMain/kotlin/org/siloserver/silo/viewmodel/RecommendationsViewModel.kt#L118-L122

**4. For You copies Home's detail-return state but not Home's reset**

The comment on L609-613 states why Home clears its token on explicit reselect (added by [`04fb276e`](https://github.com/Silo-Server/silo-android/commit/04fb276e)): a stale nonzero token "keeps suppressing Home's normal first-card focus request for the rest of the shell session." The new `forYouDetailReturnFocusRequest` / `forYouDetailReturnFocusPending` are used identically by `detailReturnFallback`, but the `ForYou` branch at L605-607 never resets them. `consumeForYouDetailReturn()` only clears on a matching `requestId`, so a restore interrupted before consumption leaves the token stuck.

https://github.com/Silo-Server/silo-android/blob/4d30facafbb7ac9a4bf12c6a5f3eb5f811f62b02/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/shell/TvMainShell.kt#L603-L619

**5. `shouldFallbackForYouReturnToFilter` is never called**

Introduced and unit-tested as the encoding of the empty-feed fallback rule, but `TvRecommendationsScreen.kt` reimplements the same `target == null` check inline. The test can keep passing while the shipped path drifts.

https://github.com/Silo-Server/silo-android/blob/4d30facafbb7ac9a4bf12c6a5f3eb5f811f62b02/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/screens/recommendations/TvRecommendationsFocusBridge.kt#L42-L50

---

Checked and clear:

- The `snapshotFlow` rewrite of the `delay(80)` polling loop from #161 is byte-identical to #162's version and preserves the "stay armed until relocation resolves" guarantee requested in [#161 (comment)](https://github.com/Silo-Server/silo-android/pull/161#discussion_r3704752922). Genuine improvement over the merged loop.
- #156's `currentRoute != TvMainRoute.Settings.route` scrim guard survives the top-bar solid/gradient refactor.
- "Supersedes #162" verified — #162 was never actually reviewed (rate limited), so no feedback was dropped in the redo.
- No persisted or phone-side state depends on the old section-id format.
- `TvRecommendationsFocusBridge.kt`'s bounded retry, stale-`requestId` rejection, and row/card requester selection check out, as do the `TvMediaRow` restore-focus additions.

Minor: every checkbox in `docs/superpowers/plans/2026-08-03-shield-focus-restoration.md` is still `- [ ]`, while the companion `2026-08-03-for-you-late-focus-relocation.md` plan has all boxes checked.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

RXWatcher:
Fixed both follow-up nitpicks in `9ee61dbb`:

- added Home-return coverage for both retry-required and immediate-success branches
- marked completed remediation-plan implementation and verification steps as complete

Validation: `./gradlew :androidTvApp:testDebugUnitTest :androidTvApp:assembleDebug` (`BUILD SUCCESSFUL`). The same patch is propagated to stacked PR #167 as `ca356792`.

RXWatcher:
Closing as superseded — everything here is already on `main`.

Checked against `main` @ b52eb877:

- the branch is **108 commits behind** `main` and **1 ahead**
- that one commit (`9ee61dbb test(tv): cover Home detail retry branches`) cherry-picks to an empty diff — `TvDetailReturnFocusStateTest.kt` is byte-identical on `main`
- `TvDetailReturnFocusState.kt` and `TvMediaRow.kt` are identical to `main`
- `TvRecommendationsFocusBridge.kt` and `TvMainShell.kt` differ only because `main` has newer commits touching them; **zero** commits from this branch are missing upstream

So the focus-restoration work landed through other PRs, and rebasing 108 commits of already-merged history would gain nothing.

Reopen or re-cut from current `main` if any of the behaviour turns out to be missing in practice.

PR #168, opened 2026-08-04
fix(tv,phone): focus ownership, return focus, and voice search

RXWatcher:
Bounded, observed TV focus acquisition; return focus by stable section and item identity rather than saved indices, which only describe the same card while the data is unchanged; modal and overlay focus ownership; centralised IME handling; and voice search from the remote.

Return restoration handles refreshes, pagination, scrolling, requester attachment and confirmed focus, across Calendar, Collections, Library, People, Personal, Requests and Search. Device pairing now requires resolved lookup data before it will offer a decision, and stale lookup and pagination results are ignored rather than allowed to land late.

## Review findings

All three CodeRabbit threads are addressed on the follow-up branch rather than here, to keep this PR at its reviewed size:

- **`HomeViewModel` fetch race** — overlapping `fetchSections()` publishers, where an older partial response could replace a newer complete one and mark sections not fully resolved. Closed by a `fetchGeneration` guard.
- **`TvSkylineSectionFeed` shared saveable slot** — `rememberSaveable` slots are positional, so two feeds composed at the same position in different surfaces shared one, and a return target armed on one could restore into the other. The feed now takes a required surface key.
- **`variant.enableUnitTest`** — deprecated in AGP 8.10.1 and removed in 9.0; replaced with the host-tests API, with the task list still offering only `testDebugUnitTest`.

> Note: this description was rewritten after the original was accidentally overwritten while folding in follow-up work. The scope and commits are unchanged; only the prose is new.

### DIFF
--- a/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/components/SiloKeyboardActions.kt
+++ b/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/components/SiloKeyboardActions.kt
@@ -1,0 +1,24 @@
+package org.siloserver.silo.android.ui.components
+
+import androidx.compose.foundation.text.KeyboardActions
+
+/**
+ * Keyboard actions for a Silo phone text field.
+ *
+ * Installing `KeyboardActions(onAny = ...)` takes ownership of the IME event.
+ * The shared field families did that unconditionally against a no-op default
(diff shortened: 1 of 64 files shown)


RXWatcher:
@coderabbitai review

RXWatcher:
@coderabbitai review

RXWatcher:
@coderabbitai review

PR #170, opened 2026-08-06
fix(player): keep playback alive on bad input, and stop sessions outliving their screen

RXWatcher:
First of three follow-ups to #168, split by theme so each stays reviewable. Player robustness; the audio-selection and profiles/navigation work follows in its own PR.

## Playback survives bad input

A malformed caption could kill playback outright. Subtitle bitmaps arriving from the wire are now rejected rather than trusted, and track identity is no longer misread — both cases previously took the whole session down rather than dropping the track.

## Sessions stop when their screen does

Server sessions could outlive the screen that started them: cancelled mid-start, superseded by a replan, or abandoned on an error path that returned before teardown. Ownership is now retained at Ready, and every session the screen adopts is owned — so it is stopped exactly once, by whoever owns it.

A frame-baseline attempt at the same problem is included here as its own commit **and its revert**. It compared decoder counters against a mount baseline, which fails because Media3 creates fresh `DecoderCounters` when a renderer is enabled — an outgoing count against a restarted counter makes a healthy stream look frozen until it has rendered as many frames again. That trades a rare missed freeze for a common false one. It is left in the history rather than squashed away because the reasoning is worth keeping.

## Overlapping refreshes

`fetchSections()` publishers could overlap — a resume observer firing while an initial load is still running — and each captured `hadSections` before its network call. An older partial response could then land after a newer complete one, replace good sections, and mark them not fully resolved, which also told TV focus restoration to keep waiting for rows that had already arrived. A `fetchGeneration` guard now serialises them: every fetch is stamped, and a result is dropped if a newer one has started.

This closes the `HomeViewModel` finding CodeRabbit raised on #168, which merged before the fix.

## Also

Audio delay, subtitle encodings and refresh reporting were unbroken; auto-play next episode was unbroken; the TV HUD's row layout, version labels and stats naming were repaired.

## Verification

All four module suites pass and both APKs assemble. Fixes carry regression tests where the boundary is testable.

## Summary by CodeRabbit

* **New Features**
  * Audio selections can carry between TV episodes when a matching track is available.
  * TV version choices now show clearer labels when versions share the same name.
* **Bug Fixes**
  * Improved playback teardown prevents duplicate or outdated sessions from stopping active playback.
  * Audio delay, subtitle decoding, and malformed subtitle graphics are handled more reliably.
  * Prevented stale home-screen refreshes from replacing newer content.
  * TV playback now surfaces subtitle, audio, and recovery failures.
  * Improved TV player layout, controls, and overlay behavior.
  * Renamed “Bitrate” to “Estimated bandwidth” for clarity.

### DIFF
--- a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackSessionLifecycle.kt
+++ b/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackSessionLifecycle.kt
@@ -616,7 +616,14 @@ class PlaybackSessionLifecycle(
                 )
             }
 
+            // The adopted id, not the published state's. Reconnecting and
+            // Failed carry no session id, so reading it from Active alone meant
+            // leaving during an outage — or after the outage timeout gave up —
+            // never told the server to stop. The transcode then ran on until it
+            // timed out, holding a stream slot the viewer had already walked
+            // away from.
             val sessionId = (_state.value as? SessionState.Active)?.session?.sessionId
+                ?: lastAdoptedSessionId
             // Fire the final snapshot regardless — even during Reconnecting we
             // want to durably record where the user was so a fresh login resumes
             // there.
@@ -745,6 +752,18 @@ class PlaybackSessionLifecycle(
                     position = pos,
                     isPaused = lastIsPaused,
                 )
+                // Re-check ownership AFTER the call. Cancelling this job is not
+                // enough to stop what follows: the network wrapper catches
+                // cancellation and hands back a NetworkError, so a reporter
+                // belonging to the previous episode carries on and acts on a
+                // reply about a session nobody is watching any more.
+                //
+                // Everything below reacts to that reply by rewriting shared
+                // state — publishing Reconnecting, restoring Active, or
+                // starting a replacement session from the CURRENT start
+                // params. Left unguarded, a late answer about episode A does
+                // all of that to episode B.
+                if (!ownsProgressReply(sess.sessionId)) continue
                 when {
                     isPlaybackSessionMissing(result) -> handleSessionMissing(sess.sessionId)
                     result is ApiResult.NetworkError -> {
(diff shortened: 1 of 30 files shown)


RXWatcher:
@coderabbitai review

RXWatcher:
@coderabbitai review

RXWatcher:
Addressed in 14e1a4bf. Six of the seven are fixed; one is a deliberate decline with reasoning.

**Fixed**

- **HomeViewModel: capture the generation before cached bootstrap** — the cached read suspends, so a refresh could publish fresh sections while we were in there and the overlay would put stale rows back. Generation is captured before the read; if it moved we skip the cached publish.
- **HomeViewModel: `isRefreshing` owned by the current refresh** — `fetchSections` now reports the generation it ran as, and only the newest clears the flag. A superseded refresh was hiding the spinner mid-fetch *and* re-opening `refreshFromRealtime`'s single-flight gate.
- **`viewModelScope.launch(NonCancellable)`** — released on the manager's cleanup scope now. Worth noting I made the new entry point *conditional*: the unconditional abandon stops even when it failed to disown, and `stopSession`'s predecessor branch then clears a newer pending publication and stops its replacement. Inline that window was small; dispatched it is not, so this fix would have opened a worse bug without the guard.
- **Three KDoc blocks before one function** — moved onto `onSubtitleFailureShown` and `onPlaybackRecoveryExhausted`.
- **Version labels resolve codec from `videoTracks`** — done, but the naive version regressed `versionPickerLabels_indistinguishableVersionsStayEqual`: two identical versions both gained the same suffix, a fabricated difference distinguishing nothing. An attribute tuple is now kept only when it actually separates something.

**Declined, with reasoning: the decoder baseline in `PlaybackStartupStallDetector`**

That baseline was implemented and reverted earlier on this same branch, in 6bd3bd55. Media3 creates fresh `DecoderCounters` when a renderer is enabled, so a baseline captured at mount can be compared against a counter that restarted at zero — the new stream would have to render as many frames again before registering as having rendered at all, and a healthy stream would sit there looking frozen until the decoder deadline fired. That trades a rare undetected freeze for a common invented one.

The comment claiming "re-baselined on arm" is what invited the suggestion, and it was wrong — only the startup deadline is cleared. It now says so plainly, records what was tried and why it went, and names the real fix: `AnalyticsListener.onRenderedFirstFrame(EventTime)` carried through a mount key on the MediaItem tag, which is device-specific Media3 integration that shouldn't be written without hardware to check it on.

**Not changed: the PGS 1080p pixel budget**

Deliberate, and the constant's comment already says so. Raising it to admit a 3840×2160 object means a ~33 MB ARGB_8888 bitmap per cue on memory-constrained TV hardware. If UHD PGS sidecars turn out to matter in practice, the right fix is downscaling at decode rather than lifting the ceiling — happy to take that as a follow-up if you've seen it drop real captions.

Full suite green on all four modules. One flake surfaced under parallel load in `PlayerViewModelLoadOwnershipIntegrationTest` (a 5s real-time `awaitCondition` deadline); it passes consistently on its own and the harness is reworked later in the stack.

PR #173, opened 2026-08-06
fix(player): address audio by the server's contract, and own the viewer's pick

RXWatcher:
Audio selection was addressed the wrong way on both clients, and the viewer's pick had no single owner.

## The server contract

The server sends **no wire `index` for audio tracks** — subtitles carry one, audio does not. Audio is addressed by its **ordinal** in the version's `audio_tracks`. Both clients were mapping picks through `AudioTrack.index`, a field the server never populates for audio, so every selection collapsed to `0`. Choosing any track played the first one.

## One owned intent

The desired track was spread across several nullable fields that raced to decide the same thing — the detail page's launch pick, a persisted fingerprint, the HUD, and the remote each wrote their own. This replaces them with one generation-owned intent that every entry point writes to, reconciled against each track snapshot by a single resolver.

That matters because a pick has to survive things that can each fail independently: staging, validation, adoption, mount, and rollback. A flag raised on *intent* carried whatever track survived a failure into the next episode as though it had been chosen; it is now set on commit.

## Direct-play streams

A direct-play stream carries every audio track, so Media3 still picked its own default even when the start request named the viewer's choice — selecting Dutch on the detail page and pressing Play mounted English. Both clients now switch on the player when the stream already carries the track, rather than assuming the server's plan settled it.

## Labelling

Player audio is labelled by **source identity** rather than delivered format. A transcoded stream reports what the transcoder produced, so a 5.1 source delivered as stereo was labelled stereo — the viewer could no longer find the track they picked.

## Shared logic

The reconcile decision and its intent types are extracted and shared between the clients, with the TV path's failure modes covered by tests, so the two implementations cannot drift apart again.

## Testing

Full suite green on all four modules. One flake surfaced under parallel load in `PlayerViewModelLoadOwnershipIntegrationTest` — a 5s real-time `awaitCondition` deadline, unrelated to this change; it passes consistently on re-run, and the harness is reworked later in the stack.

Merged `main` in to resolve one adjacency conflict in `TvPlayerViewModel` where this branch's audio block met a KDoc relocated by #170; both sides kept.

## Summary by CodeRabbit

* **New Features**
  * Improved audio-track selection and persistence across Android mobile and TV playback.
  * Added clearer audio labels, default indicators, and support for tracks unavailable in the current stream.
  * Preserved explicitly chosen audio across navigation, remounts, episode handoffs, and file changes.
  * Added reliable local switching with confirmation and recovery when tracks become available.

* **Bug Fixes**
  * Fixed incorrect audio selections caused by unreliable track identifiers.
  * Improved handling of reordered, duplicate, missing, or ambiguous audio tracks.
  * Prevented restored selections from being mistaken for newly chosen tracks.


RXWatcher:
@coderabbitai review

PR #178, opened 2026-08-06
fix(playback): make session ownership survive cancellation on both clients

RXWatcher:
Every defect here is the same shape: something proves it owns a playback session, suspends, and never re-checks — or hands ownership on without moving the token that names it. The result is a transcode nobody remembers, holding a server stream slot until it expires, or a teardown that stops the wrong session and lets the right one run on.

## PlaybackSessionManager

**An allocation lease.** After `startPlaybackV3` returns a session id, every branch still suspends before it publishes that id or stops it — and a cancellation in between left the id owned by nobody: the manager never published it, and callers only learn it when the function returns. The lease is armed the moment the response decodes and cleared only where responsibility genuinely moves.

That includes across the nested legacy-engine replan. Being installed as the active attempt is not the same as being *findable*: nothing outside the call has the id yet, and the replan suspends twice more before reaching cleanup of its own.

**Ownership released without holding the lock.** The internal replan paths clear `activeVideoAttempt` or remove the staged handle before their own suspending stop, and they run from ViewModel recovery jobs that exit and content replacement cancel. Releases now register under the caller's lock and are issued once it is released — awaiting a 60s-timeout request under `videoAttemptMutex` would serialise every start, replan and content reset behind a dying session's teardown.

Each release is tagged with a claim so one caller cannot take another's work, in-flight releases are counted rather than flagged, and the orphan ledger is bounded: only a discharged stop removes an entry, so a server that keeps failing this call grew it without limit.

## Lifecycle and both ViewModels

The ownership token is **snapshotted rather than derived** — `Reconnecting` and `Failed` deliberately keep owning their session, so reconstructing the token from the restored state erased ownership for exactly the states meant to survive an outage.

Phone exit and `onCleared` go behind the one-shot teardown gate; both clients read the retained token ahead of UI state, because every adoption path takes ownership *before* it publishes; adoption is conditional with cancellation-safe candidate cleanup; and seek recovery deliberately does **not** stop on refusal, because a re-anchor is validated to reuse the base session — that id names the session still playing.

`reportPosition` now requires the caller to name its session. `null` means "I own no session", not "skip the check": both exit paths clear the UI session id while player callbacks are still draining.

## Review

Nine Codex rounds. The last was against this base after rebasing — which is where the real defects have consistently been — and it found the replan lease boundary above. Fixing that introduced a second hazard in the same change (a successful replan would have had the `finally` stop a live session), closed in the same commit.

Full suite green on all four modules with `--rerun-tasks`. Not device-verified.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playback session stability during screen changes, recovery, seeking, and subtitle updates.
  * Prevented stale playback updates from affecting a newer active session.
  * Ensured replaced or cancelled playback sessions are stopped and cleaned up reliably.
  * Improved handling of temporary playback failures and disconnected sessions.
  * Prevented exiting one player screen from stopping playback belonging to a replacement screen.
  * Strengthened position reporting so updates apply only to the correct active session.

### DIFF
--- a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackSessionLifecycle.kt
+++ b/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackSessionLifecycle.kt
@@ -101,6 +101,15 @@ class PlaybackSessionLifecycle(
 
     private data class ActiveSessionSnapshot(
         val state: SessionState,
+        /**
+         * The ownership token as it stood, captured rather than derived.
+         *
+         * [SessionState] carries a session id only while Active, but
+         * Reconnecting and Failed deliberately keep owning theirs — so
+         * reconstructing the token from the restored state alone erases
+         * ownership exactly for the states that exist to survive an outage.
+         */
+        val lastAdoptedSessionId: String?,
         val notice: PlayerNotice?,
         val lastStartParams: StartParams?,
         val lastReportedPosition: Double?,
@@ -366,6 +375,7 @@ class PlaybackSessionLifecycle(
     private fun captureActiveSessionSnapshot(): ActiveSessionSnapshot =
         ActiveSessionSnapshot(
             state = _state.value,
+            lastAdoptedSessionId = lastAdoptedSessionId,
             notice = _notice.value,
             lastStartParams = lastStartParams,
             lastReportedPosition = lastReportedPosition,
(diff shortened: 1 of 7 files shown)


RXWatcher:
Good catch — fixed.

You're right that the capture is stale by construction: settlement can roll a subtitle publication back, and that rollback returns ownership to the predecessor, so a value read before the callback names the discarded replacement. The guard then refuses it while the one-shot gate has already been consumed, and the predecessor's session runs on until the server expires it.

`exitSessionId` is now read inside the callback. Kept the null guard as well — a screen that never owned a session has nothing to tear down, and an unqualified stop would disable the ownership check entirely.

This is the read-site counterpart to a fix already in this series: the token is moved back to the predecessor on rollback, so anything snapshotted before settlement was going to be wrong.

Full suite green on all four modules.

PR #179, opened 2026-08-06
fix(admin, tv): stop a household profile being labelled and treated as admin

RXWatcher:
Reported from a device: **a household profile that is not the owner was marked as an admin.**

That turned out to be two separate things, and the visible one was not where I first looked.

## 1. The label — what was actually on screen

The TV profile menu shows a **profile's** name with the **account's** role beneath it, so a household profile on an admin account rendered as `laura · ADMIN`. It never consulted the admin gate at all:

```kotlin
// Subtitle mirrors tvOS §5.8: role when known, falling back to username.
val subtitle = user?.role?...
```

Technically true — it is an admin account — but it reads as "laura is an admin", which is the thing that was reported.

The role now shows only on the primary profile, where it can actually be exercised. A non-owner profile shows its name and the server and nothing else: falling back to the account username just captioned laura's profile with the owner's name, conflating profile and account all over again.

No permission hangs on this label. It is the part that misleads.

## 2. The gate — what was actually exploitable

`isActingAdmin` read

```kotlin
user?.role == ADMIN_ROLE && (profile == null || profile.isPrimary)
```

so an **unresolved profile granted admin**. The account role is identical on every profile, which makes the profile the only thing separating owner from child — and Settings passes `null` explicitly when the profile lookup fails, so a load that merely errored revealed admin. The gate now requires the primary profile.

**Failing closed had to be done carefully.** Three call sites evaluate once and hold, so failing closed as first written would have hidden Admin from a genuine owner for the life of the ViewModel — worse than the bug. Both settings ViewModels now retry an unresolved profile, **bounded**: not being an admin is by far the commonest reason for no admin row, and an earlier attempt that retried on "admin not visible" would have looped on the API forever for every ordinary user.

**Gating the entry was not enough.** The phone admin route stays registered and its screen calls the API as it composes, so restored or direct navigation reached it regardless of the menu row. `AdminRouteGate` re-evaluates at the destination and refuses, rendering nothing while the gate resolves.

## Security impact

The server **does** enforce this. `RequireActingAdmin` checks the admin role *and* that the declared profile is the account's primary, and fails closed on a profile it cannot resolve — explicitly so a non-primary session cannot regain admin with a bogus `X-Profile-Id`.

So the reported incident was **UI exposure, not privilege escalation**.

One narrow exception, worth recording: the server allows admin on role alone when **no** profile is declared, and the client omits `X-Profile-Id` when the profile id is null — the same unresolved-profile condition. In that window a non-primary profile on an admin account could have reached admin data. This PR closes it client-side at both the entry and the destination.

The server is also candid that this is a policy boundary for well-behaved clients, not isolation between people sharing an account.

## Also

A test helper ignored both its arguments and always returned `true`, so every test through it passed regardless of the gate — including the case the class exists for. Fixed, with coverage for an admin account on a non-primary profile, on an unresolved profile, and on the primary profile.

## Testing

Full suite green on all four modules. Verified on a Google TV Streamer with a release build: the profile menu now reads `laura · WAVE-NINJA` instead of `laura · ADMIN · WAVE-NINJA`.

Reviewed by Codex, which caught the latching, the ungated destinations and the overclaiming comments. The label fix came out of looking at the actual screen — no amount of gate review would have found it, because the label never touched the gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved admin access checks by requiring a resolved primary profile.
  * Added bounded retries for temporary profile-resolution failures.
  * Admin areas now display loading, authorized, or access-denied states appropriately.
  * Prevented unauthorized access when profile information is unavailable.
  * Updated TV profile details to show account roles only for primary profiles.

* **Tests**
  * Added coverage for denied access, unresolved profiles, and recovery after temporary lookup failures.

### DIFF
--- a/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/navigation/AppNavigation.kt
+++ b/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/navigation/AppNavigation.kt
@@ -1066,9 +1066,14 @@ fun AppNavigation(
             )
         }
         composable(Route.Admin.route) {
-            org.siloserver.silo.android.ui.screens.admin.AdminStatsScreen(
-                onBackClick = { navController.popBackStack() },
-            )
+            // Gated at the destination as well as the entry: the route stays
+            // registered, so restored navigation reaches it directly and the
+            // stats screen calls the admin API the moment it composes.
+            org.siloserver.silo.android.ui.screens.admin.AdminRouteGate {
+                org.siloserver.silo.android.ui.screens.admin.AdminStatsScreen(
+                    onBackClick = { navController.popBackStack() },
+                )
+            }
         }
         composable(Route.Watchlist.route) {
             WatchlistScreen(
--- a/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/admin/AdminEntryViewModel.kt
+++ b/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/screens/admin/AdminEntryViewModel.kt
@@ -7,6 +7,7 @@ import org.siloserver.silo.model.auth.isActingAdmin
 import org.siloserver.silo.network.ApiResult
 import org.siloserver.silo.repository.AuthRepository
 import org.siloserver.silo.repository.ProfileRepository
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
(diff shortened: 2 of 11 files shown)


RXWatcher:
@coderabbitai review

RXWatcher:
Fixed — and a fair catch, since you quoted my own KDoc back at me.

I documented on `isActingAdmin` that call sites must retry or observe because the gate fails closed, fixed the two settings ViewModels accordingly, and then added a brand-new destination gate on the third ViewModel without applying the same rule. So the gate introduced to close a hole could itself strand the owner it exists for — and on a restored or directly-navigated route, for the whole life of that entry.

`AdminEntryViewModel`'s repo-backed provider now retries an unresolved profile, bounded, matching the settings ViewModels. `getActiveProfile` collapses "network failed", "no active id" and "not found" into null, so a retry is the only signal available. Bounded because not being an admin is the ordinary case, and an unbounded retry would poll for every non-admin who reaches this route.

Test added for unresolved → primary while the destination is still active, as suggested.

Full suite green on all four modules.

PR #183, opened 2026-08-06
fix(tv): stop a For You detail return leaving the screen without focus

RXWatcher:
Returning from a detail screen could leave For You with **no focus owner at all**, so the D-pad stopped responding until you navigated away.

## The dead end

When the relocation loop cannot reach the launch card it reports `Exhausted`. The entire recovery for that case was one unchecked call:

```kotlin
if (result == ForYouReturnFocusResult.Exhausted) {
    requestFocusSafely { forYouFocusRequester.requestFocus() }
}
latestOnDetailReturnFocusConsumed(detailReturnFocusRequest)
```

`requestFocusSafely` returns `Handled` / `Rejected` / `Disposed`, and the result was discarded. If that request did not take — the pill not attached on the frame we happened to ask — nothing retried, no other candidate was tried, and nothing was logged.

The silence is the reason this was hard to find. `requestFocusSafely` deliberately converts the `FocusRequester is not initialized` throw into a value rather than letting it surface, so a failed last-resort claim produced no warning in logcat at all.

The no-target branch a few lines above already does the right thing — it retries `forYouFocusRequester` across frames and only gives up after `TvFrameRelocationMaxAttempts`. This path just never got the same treatment.

## The fix

Retry across frames, and fall through the filter pills (`For You` / `Watchlist` / `Favorites`). Those are composed for the life of the screen, so one can take focus even while the feed is still settling. A candidate that throws is treated as *not yet attached* rather than fatal, since the node can attach on a later frame. If nothing takes focus, say so.

## Why it matches the report

Reported as intermittent, roughly one in three, on Shield after `Watchlist → item → Play → Back`:

- **intermittent** — needs the relocation loop to exhaust *and* the fallback to miss
- **Shield-specific** — slower device, more frames before the row attaches
- **after playback specifically** — the most disruptive recomposition on that screen
- **genuinely stuck rather than misplaced** — focus lands nowhere, not on the wrong element

## Verification

- new `ForYouFallbackFocusTest` covers first-candidate claim, fall-through, later-frame attachment, a throwing candidate, total failure, and no candidates
- `:androidTvApp:testDebugUnitTest` green (forced rerun)
- `:androidTvApp:compileDebugKotlinAndroid` clean

**Not reproduced directly.** Four cycles on a Google TV Streamer stayed clean — this was found by reading the path, not by catching it live. Worth a Shield test before merging.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved focus restoration when returning from content details.
  * Added fallback focus handling across For You, Watchlist, and Favorites controls.
  * Focus recovery now retries safely across frames and skips unavailable controls.
* **Tests**
  * Added coverage for successful recovery, fallback selection, retries, exceptions, exhausted candidates, and empty lists.


RXWatcher:
@coderabbitai review

RXWatcher:
@coderabbitai full review

RXWatcher:
@coderabbitai review

PR #184, opened 2026-08-06
fix(tv): let the playback selector menus reach every option

RXWatcher:
## Problem

On the TV detail screen, subtitle selection stopped working past the tenth row. D-pad Down walked to the last row that happened to be on screen, then focus left the popup for the detail screen behind it, and the list never scrolled. On a title with 17 subtitle tracks, everything from the seventh onwards was unreachable.

Reported against `v1.0.0-rc.3+6`. The previous fix for this (`46f60eef`, which *is* `v1.0.0-rc.2+5`) has been shipping in two release candidates without effect.

## Root cause

Measured on a Google TV Streamer. Two independent faults:

1. **Focus leaves the popup.** Compose's focus search abandons the `DropdownMenu` once the next row is off-screen — focus was observed landing on the detail screen's Resume button while the menu was still open. The rows that needed scrolling were never reached.
2. **`BringIntoViewRequester` does not scroll this menu.** An injected scroll moves the same list end to end, so the container is scrollable; the requester simply has no effect inside the popup.

The rc.2 fix addressed only the second, using the mechanism that does not work here. It could only ever scroll a row that focus reached, and focus could not reach the rows that needed scrolling.

## Approach

The menu now owns both halves of the walk:

- `nextSelectorMenuIndex` drives d-pad movement, stepping over disabled rows (the "Unknown" audio fallback) and consuming at the boundaries so focus cannot leak to the screen underneath.
- `selectorMenuScrollTarget` scrolls the menu's own `ScrollState` from measured row positions. `BringIntoViewRequester` is removed.

## Verification

On a Google TV Streamer against a 17-track title:

- focus walks the visible rows with the list static (nothing to scroll)
- from the last visible row, the list scrolls one row per press to the end
- Up scrolls back to the top
- both ends hold; focus never leaves the menu

9 unit tests cover the walk and the scroll arithmetic; the `androidTvApp` suite is green.

## Scope

`Version`, `Audio` and `Edition` share this component and get the same fix. No server, API or model changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TV menu navigation with reliable focus when opening.
  * Up and Down navigation now skips disabled options and stops at menu boundaries.
  * Automatically scrolls to keep the focused option visible.
  * Added fallback focus to the first selectable option when no selection exists.

* **Tests**
  * Added coverage for navigation, disabled options, focus behavior, scrolling, boundaries, and unmeasured layouts.

### DIFF
--- a/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/components/TvAnchoredSelectorMenu.kt
+++ b/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/components/TvAnchoredSelectorMenu.kt
@@ -2,9 +2,11 @@ package org.siloserver.silo.tv.ui.components
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.border
+import androidx.compose.foundation.focusGroup
 import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.interaction.collectIsFocusedAsState
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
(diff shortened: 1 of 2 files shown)


RXWatcher:
Good catch — fixed in `aa6fdc98`.

`initialSelectorMenuIndex` coerced the no-enabled-row result to `0`, which is a *disabled* row. It can't take focus, so the request failed silently and the menu opened with focus nowhere and the d-pad dead.

It now returns `-1` for that case and the focus request is skipped. `nextSelectorMenuIndex` already handles starting from `-1` (it steps to candidate 0 and walks forward), so a d-pad press still reaches the first selectable row. Tests added for the all-disabled list, the empty list, and the walk out of the `-1` state.

PR #185, opened 2026-08-06
perf(tv): split the player overlays out of TvPlayerScreen

RXWatcher:
## Problem

`TvPlayerScreen` had grown to 2083 lines, and ART refused to compile the composable it generates:

```
Method exceeds compiler instruction limit: 17433 in void
TvPlayerScreenKt.TvPlayerScreen(...)
```

Past that limit the method is never JIT-compiled, so the entire player screen runs interpreted. On a Google TV Streamer the runtime logged this roughly once a second for the whole duration of playback.

## Approach

The overlay layer is the most self-contained part of the composable, and it is UI rather than effects — so unlike the `LaunchedEffect` blocks (whose lambda bodies already compile to their own methods), its call sites carry parameter setup and change-detection code inline in the parent method.

`TvPlayerOverlays` takes with it:

- the lifecycle notice toast
- the remote-control `display_message` toast
- the Watch Together room indicator and host close-confirm dialog
- the Up Next end-of-playback surface
- the intro auto-skip banner
- the reconnect spinner

It takes plain values and callbacks and holds no state of its own, so the mutable state (`showLeaveDialog`) and the local helpers (`stopPlaybackAndExit`, `handleSkipIntroNow`) stay in `TvPlayerScreen` and are passed in.

## Verification

On a Google TV Streamer during live playback (`isPlaying: true`, position advancing), the message no longer appears at all — before the change it fired continuously. `androidTvApp` unit suite green.

## Scope

Pure extraction: no behaviour change, no server/API/model changes. Worth noting the composable is still large and other seams remain (the 24 `LaunchedEffect`s and 10 `DisposableEffect`s, and the main UI tree); this change is what was needed to get under the limit, not a full decomposition.

### DIFF
--- a/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/screens/player/TvPlayerScreen.kt
+++ b/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/screens/player/TvPlayerScreen.kt
@@ -90,6 +90,7 @@ import org.siloserver.silo.common.player.SiloPlaybackService
 import org.siloserver.silo.common.player.DisplayHdrProbe
 import org.siloserver.silo.common.player.HdrDisplayController
 import org.siloserver.silo.common.player.PlaybackCapabilityDetector
+import org.siloserver.silo.common.player.PlayerNotice
 import org.siloserver.silo.common.player.PlaybackPreflightListener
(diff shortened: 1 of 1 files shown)


RXWatcher:
@coderabbitai review

RXWatcher:
@coderabbitai review

PR #187, opened 2026-08-06
fix(playback): let the audio route settle before reselecting tracks

RXWatcher:
## Problem

A KVM switching inputs killed playback outright:

```
NullPointerException: MediaPeriodHolder.info
  at ExoPlayerImplInternal.seekToCurrentPosition(...:1225)
  at ExoPlayerImplInternal.reselectTracksInternalAndSeek(...:2303)
```

A reselection is resolved by seeking the current media period. While an audio sink is being torn down and rebuilt there is no such period, so the seek dereferences a null holder. Pulling HDMI produces exactly that: the sink is rebuilt and capabilities are re-reported mid-rebuild, and the `LaunchedEffect` keyed on those capabilities re-applies presets straight into the gap.

## Why the previous attempts could not work

Both guarded the call site by sampling the player:

- **#182** — `playbackState` and `currentTimeline.isEmpty`
- **#186** — those, plus session state and a no-op parameter check

Neither can see the window. The player looks healthy from the caller's thread throughout, because the teardown completes on ExoPlayer's own. #182 shipped through two release candidates without effect; #186 reproduced the crash on the very build that contained it, with all its guards passing legitimately:

| guard | value at crash |
| --- | --- |
| `exitRequested` | false — playback live |
| `sessionId == null` | false — session healthy |
| `next == base` | false — capabilities genuinely changed |

## Approach

Stop trying to observe the window; stay out of it.

A capability change now waits before presets are re-applied. Because the effect restarts on every capability report, route churn coalesces into a single application once the reports stop. The first application for a session still runs immediately, so startup track selection is not deferred.

## Verification

On a Google TV Streamer, across a KVM switching away and back repeatedly:

| | |
| --- | --- |
| HDMI route changes | **8** (add/remove/add) |
| including rapid churn | remove → add → remove within ~2s — the pattern that reproduced the crash |
| `seekToCurrentPosition` NPEs | **0** |
| player errors | **0** |
| session | survived |

The previous crash coincided exactly with an `addProviderRoutes … ROUTE_ID_HDMI`; that transition occurred three times here without incident.

Unit suites green across `android-shared`, `androidTvApp`, `androidApp`.

**Limits of this result:** one session, and playback was paused by the end, so not every switch is confirmed mid-play. The mechanism is play-state independent — the effect fires on capability change either way — but a second run playing throughout would make it airtight.

## Scope

The phone screen carries the same effect, reached by a headphone unplug or Bluetooth drop rather than an HDMI switch, and gets the same treatment. Supersedes #186 (closed).

### DIFF
--- a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/SiloPlayerFactory.kt
+++ b/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/SiloPlayerFactory.kt
@@ -375,6 +375,11 @@ class SiloPlayerFactory(
      * Apply capability-aware track selection presets to [player]. Call at
      * player construction and again whenever [AudioPassthroughCapabilities]
      * changes (HDMI hot-plug, BT pair, user-toggled "force Atmos" setting).
+     *
+     * Returns whether parameters were actually assigned. A skip — teardown
+     * guard or no-op parameters — must be visible to callers that track
+     * "presets have been applied once", or a skipped first run silently
+     * reclassifies the real first application as a later change.
      */
     fun applyTrackSelectionPresets(
         player: Player,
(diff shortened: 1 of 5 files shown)


RXWatcher:
@coderabbitai review

RXWatcher:
@coderabbitai review

PR #188, opened 2026-08-06
fix(tv): stop a duplicate feed entry crashing the app

RXWatcher:
## Problem

Reported from a release build — a **fatal** crash, the app dies:

```
IllegalArgumentException: Key "series-tvdb-280619" was already used.
If you are using LazyColumn/Row please make sure you provide a unique key for each item.
```

Compose requires lazy-list keys to be unique and throws on the draw pass when they are not. A feed that returns the same series twice in one row therefore takes the whole app down. No feed response should be able to do that.

## Approach

Three shared TV surfaces key their cards on `contentId` — `TvMediaRow`, `TvCatalogGrid` and `TvHomeHeroCarousel` — and all three now deduplicate before keying.

This is correct on its own terms, not just as a crash guard: a row has no reason to draw the same title twice. Feeds legitimately overlap — a title can sit in Continue Watching *and* in a genre row — so uniqueness is a property each list should hold for itself rather than something every caller must guarantee.

The deduplicated list is then used for **everything that indexes into it**, not only for rendering:

- the hero carousel reports its active item as `items[activeIndex]` and auto-advances modulo the list size
- the grid compares the last visible index against the list size to trigger load-more

Leaving those pointed at the caller's list would have traded a crash for a carousel that reports a different card than the one it drew.

## Verification

`androidTvApp` unit suite green; TV app compiles.

**Not reproduced locally** — the report is from someone else's device and the stack is obfuscated, so I can't say which of the three surfaces threw. All three share the shape, so all three are fixed.

## Follow-ups (not in this PR)

- **Why is there a duplicate at all?** A row showing one series twice is wrong even without the crash. Diagnosing it needs the feed that produced it.
- **The phone app keys 8 lists on `contentId`** the same way and has the same exposure. Each needs the same index-desync check performed here, and the reported crash is on TV, so they are left alone rather than changed blind.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed duplicate titles from TV catalog grids, home carousels, and media rows.
  * Improved pagination, scrolling, auto-advance, and page indicators when duplicate content is received.
  * Prevented repeated loading attempts and display errors caused by duplicate items.

### DIFF
--- a/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/components/TvCatalogGrid.kt
+++ b/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/components/TvCatalogGrid.kt
@@ -116,6 +116,10 @@ fun TvCatalogGrid(
     emptyState: (@Composable () -> Unit)? = null,
 ) {
     val resolvedGridState = gridState ?: rememberLazyGridState()
+    // Keyed lazy lists throw on a repeated key, which is fatal. Paging can
+    // hand the same item back across pages, so the grid guarantees uniqueness
+    // itself rather than trusting every caller to.
+    val uniqueItems = remember(items) { items.distinctBy { it.contentId } }
 
     // Backoff gate against an endless load-more retry storm. When a load-more
     // completes without growing the list while the server still reports more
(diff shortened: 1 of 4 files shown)


RXWatcher:
Thanks — one of these was a real bug I introduced, one is pre-existing.

**Focus restoration desync — fixed** (`aeea575a`). Correct catch. `TvMediaRow` renders the deduplicated list while `restoreFocusIndex` counts positions in the caller's, which `TvSkylineSectionFeed` computes with `section.items.indexOfFirst { … }`. A duplicate earlier in a row shifted every position after it, so focus restored to the wrong card — trading a crash for a subtler bug.

The row now translates the incoming index through `contentId` into its own list. It also reports the **caller's** index back through `onRestoreFocusTargetDisposed` rather than its own, because `TvRecommendationsScreen` compares that against the value it passed in — the same coordinate-space mismatch in the other direction.

**Pagination reset keying — not changing here.** `main` already keys that effect on `items.firstOrNull()?.contentId, items.size`; this PR only swapped which list it reads. Keying on the full ordered sequence may well be an improvement, but it's pre-existing behaviour and unrelated to the crash this PR fixes, so it belongs in its own change rather than riding along.

PR #196, opened 2026-08-07
chore: gate API-level mistakes with lint in CI

RXWatcher:
Follow-up to #195. Two startup crashes reached users this week because nothing
in this project checked API levels — lint had never been configured, had no
baseline, and did not run in CI.

## Three API-guard bugs, all in `AudioCapabilityManager`

| Call | Level | Effect |
| --- | --- | --- |
| `AudioDeviceInfo.getAddress()` | API 28, **no guard at all** | `NoSuchMethodError` on API 24–27 whenever diagnostics were collected |
| `AudioTrack.isDirectPlaybackSupported()` | API 29, only gated for TIRAMISU | `runCatching` swallowed the resulting error on API 24–28, so **"no passthrough" was an accident rather than a decision** |
| `AudioFormat.ENCODING_E_AC3_JOC` | API 28, **no `minSdk`** | Queried on API 24–27 for an encoding that does not exist there |

The first is the same shape as #195 and is a real behaviour fix.

The second is **not** a behaviour change, and the table above previously implied
it was. Pre-Q devices reported no passthrough before this change and still do —
the platform offers no way to ask before API 29, so `false` is the only honest
answer. What changes is that the answer is now an explicit branch instead of a
swallowed `NoSuchMethodError`, which means lint can see it and a reader can tell
it was intended. Worth having, but nobody's audio improves.

The third was found by this PR's own lint run. `EncodingSupport` carries a
`minSdk` field so a constant is only queried where it exists — `dts_hd`,
`truehd` and `ac4` all use it. `eac3_joc` needs API 28 exactly as `ac4` does
and was the one entry left on the default of `1`. It cannot crash (the
constant inlines, and the platform returns false for a value it does not
know), so the result was right by accident in a table where every neighbour
is right by design.

Note that this does **not** remove a baseline entry: `InlinedApi` fires on the
constant reference, not the runtime guard, which is why correctly-guarded
`truehd` and `ac4` are baselined too. All three entries stay. The fix is about
the code being right, not about silencing lint.

## The gate

- `NewApi` and `InlinedApi` are **fatal** on `android-shared`, `androidApp` and
  `androidTvApp`.
- `checkDependencies = true` on both apps. This is the structural part:
  `android-shared` is where both crashes lived, and app-level lint does not
  analyse library sources without it — so the module holding the risk was the
  one nothing examined.
- Per-module `lint-baseline.xml` holds today's findings so only **new**
  violations fail. They are overwhelmingly desugared `java.*` calls (core
  library desugaring is enabled; lint does not model it) and deliberate media3
  `@UnstableApi` usage.
- A `Lint` job in `android-build.yml`, mirroring the unit-test job, uploading
  reports on failure.
- The job runs `lintVitalRelease` for both apps as well. `checkReleaseBuilds`
  makes that part of `assembleRelease`, which runs only from `release.yml` on a
  `v*` tag — so the release variant was gated by a check no pull request
  exercised, and its first run would have been mid-release. It also covers the
  release-only source sets `lintDebug` never analyses.

## Verification

The two fixes were made **before** generating the baselines, so neither is
baselined — grep confirms no `Spatializer`, `getAddress` or
`isDirectPlaybackSupported` entries in any of the three files. Baselining first
would have permanently accepted both.

Reintroducing the unguarded `getAddress()` call fails the build with
`Call requires API level 28 (current min is 24)`; with the fix in place lint
reports `no new issues (242 errors … filtered by baseline)`.

## Note for future contributors

If the lint job fails, fix the call. Regenerating a baseline to make it pass is
how this protection gets quietly dismantled.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with older Android versions by preventing unsupported audio APIs from being called.
  * Refined audio capability detection for device routes and newer audio formats.

* **Quality Improvements**
  * Added Android lint checks across mobile and TV builds.
  * Release builds now enforce critical API compatibility checks and fail when lint errors are detected.
  * Lint reports are available for troubleshooting when checks fail.

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -63,3 +63,48 @@ jobs:
           path: |
             **/build/reports/tests/**
           if-no-files-found: ignore
+
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out sources
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
+        with:
+          persist-credentials: false
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95 # v5
+        with:
+          distribution: temurin
+          java-version: "21"
+
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6
+
+      - name: Run lint
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          # NewApi and InlinedApi are fatal here. Two startup crashes reached
+          # users because nothing checked API levels: a Spatializer call gated
+          # at 31 for a class introduced at 32, and an unguarded API 28 call.
+          # Existing findings are held in per-module lint-baseline.xml, so only
+          # NEW violations fail. If this job fails, fix the call — do not
+          # regenerate the baseline to make it pass.
+          ./gradlew \
+            -Dorg.gradle.jvmargs="-Xmx4g -Dfile.encoding=UTF-8" \
+            :android-shared:lintDebug :androidApp:lintDebug :androidTvApp:lintDebug \
+            --max-workers=2
+
+      - name: Upload lint reports
+        if: failure()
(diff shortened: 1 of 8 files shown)


RXWatcher:
@coderabbitai review

PR #203, opened 2026-08-10
fix(diagnostics): stop the focus warning crashing, and name the endpoints it reports

RXWatcher:
## 1. #199's new warning crashes debug builds the moment it fires

#199 added `DiagnosticsFocusLogger.contentEntryFailed(route)`, which reports a `route` attribute in the `FOCUS` category. The `FOCUS` registry in `SiloLog.kt` declares only `target` and `action`, and an unregistered attribute hits:

```kotlin
if (strictAttributeRegistry) {          // = BuildConfig.DEBUG
    throw IllegalArgumentException("Unregistered or mismatched diagnostics attribute …")
}
return@forEach                          // release: silently dropped
```

So in **debug** builds the warning added to make silent focus failures visible throws the instant it fires — and it fires exactly when TV content focus entry fails, which is the condition #199 exists to diagnose. In **release** builds it does not crash, but `route` is dropped, shipping the warning without the one field that makes it actionable.

Fix: register `route`.

## 2. An allowlisted resource now keeps its name

`safeDiagnosticsNetworkPath` collapsed any path whose *tail* matched no template down to a bare `/api/v1/other`. A tester's device produced 52 404s in an eight-minute session, all logged that way — the largest error signal on the device, completely unactionable.

An allowlisted resource now keeps its name: `/api/v1/playback/other` rather than `/api/v1/other`. That resource already appears in every other path logged for it, so nothing new is disclosed. An **unrecognised** resource stays anonymous, deliberately — the existing test makes the point with `/api/v1/private/private-id` → `/api/v1/other`, and that is unchanged, as is the query-string case.

## 3. Seven API roots had no route template at all

`favorites`, `history`, `library-playback-prefs`, `metadata`, `onboarding`, `watch`, `watchlist` — all called by the client, none in `API_ROUTE_TEMPLATES`, so all bucketed.

Two of them carry the loudest 4xx the client emits: `checkFavorite` and `checkWatchlist` answer "not in the list" with a 404, and the TV detail screen fires one per episode on every season load. On a live device those 404s were **116 of the 121 four-hundreds recorded**. (The N+1 itself is a separate fix; this is about being able to see it.)

Derived by diffing the paths the client builds under `shared/.../network` against the template keys, so the gap is closed as a set rather than one report at a time. Stale keys in the other direction (`stream`, `users`, `watch-together`) are left alone — they cost nothing and the routes still exist server-side.

## Verification

`:android-shared:testDebugUnitTest` — 1101 tests, 0 failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic data accuracy for API activity involving favorites, history, library playback settings, metadata, onboarding, watching, and watchlists.
  * Recognized resource names are now retained in diagnostic routes, while unrecognized resources continue to be anonymized.
  * Improved handling of item-specific routes by representing identifiers consistently without exposing sensitive values.
  * Added focus-related diagnostic information for more complete activity reporting.


RXWatcher:
@coderabbitai review

RXWatcher:
@coderabbitai full review

RXWatcher:
@coderabbitai full review

PR #207, opened 2026-08-10
docs: design an explicit focus-ownership model for the TV shell

RXWatcher:
Design only — no code changes. Opening it as a PR so the argument can be
disagreed with before anyone refactors the shell.

## The problem

Focus fixes are accelerating, not converging.

| Month | focus-titled commits on `main` |
|---|---|
| 2026-06 | 8 |
| 2026-07 | 26 |
| 2026-08 (to the 10th) | **49** |

Of 83 in the last 90 days, **49 are `fix:` against 9 `feat:`**. `TvMainShell.kt` has been edited by a focus commit **15 times**. There are already 19 production focus files, 22 focus test files and 8 focus design docs — the infrastructure is not missing.

## Why the last design did not stop it

`2026-08-04-whole-application-focus-hardening-design.md` was largely right and this keeps its principles. But it says:

> *Shared helpers encode repeated policy, but screen-specific resolution remains close to the screen. This avoids a global focus coordinator with hidden cross-route coupling.*

Reasonable — but it leaves the **shell** unmodelled, and the shell is where per-screen policies collide. It also has no enforcement half, and its escape clause (*"existing specialized flows … remain intact unless they can adopt the helper"*) makes divergence compliant.

Two defects found since are not among its six causes, and neither is an acquisition bug:

- **#202** — a `focusRestorer` on an ancestor intercepts focus *entry*, so no subtree-level fix can win.
- **#204** — `menuFocusTarget` meant both "which bar element" and "suppress its dwell".

Both are **ownership** bugs.

## Root cause

`TvShellFocusState` carries ten mutable fields and **none of them says who owns focus**. Ownership is inferred by reading several together; every combination is representable, including the meaningless ones. Each fix adds a flag excluding one bad combination, which enlarges the space the next defect hides in.

The insight is already in `b44e1d8f`'s commit message:

> *The distinction that matters is not which call site but **why focus arrived**.*

Correct — and that reason is stored nowhere, so it is reconstructed at each decision point, and each reconstruction is a chance to get it wrong.

## Proposal

One value: an owner (`Content` / `Bar` / `Panel` / `ProfileMenu` / `Modal`) plus **why focus arrived** there. Back routing and dwell eligibility become total functions over it — exhaustively testable without a Compose runtime, and adding an owner or reason **breaks compilation at every decision point**. That compiler enforcement is what the previous design lacked.

#204's two new flags collapse into it: `barFocusFromPanelClose` becomes `Bar + PanelDismissed`, and `menuFocusSuppressesDwell` stops existing. The bug where an ordinary Up armed Back-close suppression becomes *unrepresentable* rather than excluded by a flag.

## Scope

**In:** `TvShellFocusState`, `TvMainShell`, `TvTopMenuBar`, Back/dwell decisions — the 15-edit hotspot.
**Out:** screen-local initial-focus acquisition and the bounded observed-focus policy. That half works and stays.
**Not** a global focus coordinator — screens keep resolving their own targets; the shell stops guessing what state it is in.

## Success criteria

Stated in the doc and measurable: focus `fix:` commits per month, edits to `TvMainShell.kt` per month, and `focus trap suspected` / `navigation struggle` per tester session. If they do not fall, the document was wrong and should be replaced rather than supplemented.

Depends on #204 landing first.

### DIFF
--- a/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/screens/detail/TvItemDetailViewModel.kt
+++ b/androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/screens/detail/TvItemDetailViewModel.kt
@@ -51,6 +51,8 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Semaphore
+import kotlinx.coroutines.sync.withPermit
 
 data class TvItemDetailUiState(
     val isLoading: Boolean = true,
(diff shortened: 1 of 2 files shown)


RXWatcher:
Superseded by #211.

This was stacked on the pre-merge version of #206 and carried its original commit `ad0d92a0`, which has since gained its test suite, per-answer publication and targeted revalidation. Merging it as-is would have dragged the untested version back in.

#211 is the same design note rebuilt on merged `main` (docs-only), with one correction: it described `barFocusFromPanelClose` as a flag #204 adds, but #204 deleted it as unreachable from production UI.

PR #222, opened 2026-08-12
fix(tv): show one buffering indicator, even when the controls are hidden + reconnect verbiage change

evulhotdog:
### The wording of the reconnect banner
It used to read `Reconnecting. The server is updating. Playback will resume when it is ready.` That names one
specific cause, and the code path it comes from fires on 502/503/504 and
520-527/530 gateway errors and on plain network errors, so the real cause is
just as likely to be a crashed or restarting server, a dead tunnel or reverse
proxy, or the viewer's own network dropping. Telling someone their server is
updating when it actually fell over sends them looking in the wrong place. It
now reads `Reconnecting. Playback will resume automatically.` No guess at the
cause, and no promise about server readiness the client cannot actually check.
The matching timeout message was already worded this way, so this brings the
pair in line.

### Fix double loading spinners in the TV player

Playing a video on Android TV could put two loading indicators on screen at
the same time: a teal ring in the middle of the picture and a "Buffering"
pill in the top right corner. This branch gets it down to one indicator per
situation, and fixes a few things that fell out of doing that.

## Confirmation

The "Buffering" pill was added a while back specifically to replace the big
centered spinner, but the line that turns the centered spinner off was never
written. ExoPlayer's `PlayerView` draws its own buffering spinner when you ask
it to, and we were still asking. So both drew at once. The code comments next
to the pill even claimed it had replaced the centered spinner, which made the
leftover easy to miss on review.

## Summary by CodeRabbit

- **New Features**
  - Added a buffering indicator that remains visible when playback controls are hidden.
  - Reserved the centered spinner for reconnecting playback states.

- **Bug Fixes**
  - Updated the reconnecting message to indicate that playback resumes automatically.
  - Hid buffering and sleep-timer indicators during HUD, Up Next, picture-in-picture, and reconnect states.
  - Limited status indicators to active video playback and removed the duplicate built-in buffering spinner.


Quick104:
Pushed 430cd1b addressing the review findings on the TV half of this change. The reconnect-message reword was already clean.

**The capsule now stays up wherever video is playing.** Turning off `PlayerView`'s spinner (`SHOW_BUFFERING_NEVER`) made the capsule the only buffering feedback there is, so gating it on `!hudOpen && !showNextUp` left those two surfaces with nothing. Neither owns a loading state to fall back on:

- A HUD Quality or Version pick restarts the whole session with the HUD still open — `closeOnSelect = true` closes only the picker — which is the longest rebuffer in the app.
- Up Next appears pre-credits with video still playing behind the mini-player frame.

The capsule survives both now. To avoid colliding with the HUD's top-center card (≤680dp wide, ≤360dp tall at top 56dp — its right edge runs under the chip's 80dp end inset at typical TV widths), it drops to a 440dp top inset while the HUD is open, and takes `zIndex(5f)` so it clears the Up Next scrim. The sleep-timer chip keeps yielding the corner to both, since an ambient countdown isn't worth competing for that space.

**The chip block is now gated on `videoActive`.** Moving it out of the `state.streamUrl != null` branch into `TvPlayerOverlays` put it over `TvLoadingScreen` and `TvErrorScreen` too, and `fail()` sets `error` without clearing `isBuffering` — so a spinning "Buffering" capsule could land on the error screen. `videoActive` (`streamUrl != null && !isLoading && error == null`) is hoisted from the PiP state block, which already computed it, and passed in. Version switches use the replacement-load path that keeps `isLoading = false`, and quality switches go through the subtitle-transaction path rather than `loadContent`, so the gate holds through exactly the restarts the first fix is for.

Also corrected the two comments that asserted the HUD and Up Next own their own loading feedback.

Verified: `:androidTvApp:assembleDebug` and `./gradlew test` both pass. One flake on the first test run — `ServerSetupPersistenceTest` teardown hitting `Dispatchers.Main is used concurrently with setting it`, a test-dispatcher race in the phone auth suite, unrelated to this change and green on rerun.

PR #232, opened 2026-08-15
feat(diagnostics): re-vendor expanded attribute registry and align hosted allowlist

Quick104:
## Why

Re-vendors the attribute registry from silo-server so this client stays in lockstep with the collector. Depends on Silo-Server/silo-server#649.

Coordinating that change surfaced three gaps on the Android side, all fixed here.

## 1. `REGISTERED_ATTRIBUTES` had fallen behind

`DiagnosticsValidation.kt` still held the pre-change key set while its sibling fixture was current.

Impact was limited but real: `validate()` hits a `null -> Unit` branch for an unknown key, so those attributes lost *type validation* silently. Nothing was dropped or rejected — the emission gate that matters (`SiloLog.kt` `ATTRIBUTE_REGISTRY`) was correct. Contract-validation decay rather than data loss.

## 2. No parity gate

Nothing in the test suite read the vendored `attr-registry.json` — zero code references. Go has `TestAttrRegistryStaysInSync` and Swift now has an equivalent; Android had a dead fixture, so the exact hand-drift this whole effort exists to eliminate could still happen here silently.

Adds `DiagnosticsAttributeRegistryParityTest`, asserting `REGISTERED_ATTRIBUTES` matches the fixture in both directions including value types.

## 3. Hosted allowlist forwarded keys Apple withholds

This is the one worth a close look. `HOSTED_V1_LOG_ATTRIBUTES` was forwarding `playback.{session_id, play_method, reason, position_ms}` and `network.attempt` to the hosted collector.

`session_id` and `attempt` are in the collector's `FORBIDDEN_KEYS`, so those bundles would fail the `privacy_fields` check. The Apple client deliberately withholds all five — so as written, the two clients encoded **contradictory answers to the same privacy question** in one coordinated change.

No production Android code emits these keys yet, so this is a latent-leak fix, not a behavior change. But the first engineer to instrument Android playback would have inherited it.

Now aligned with Apple, with a test pinning the withheld set in both directions.

**Subtlety worth verifying in review:** `reason` exists in *both* `playback` and `lifecycle`. Only `playback.reason` is withheld — it's server-authored free text. `lifecycle.reason` is a client-side classification and is still forwarded. The filter is per-category so they can't collide structurally, and the test asserts both halves.

## Testing

    ANDROID_HOME=~/Library/Android/sdk ./gradlew :android-shared:testDebugUnitTest --tests "*DiagnosticsBundleBuilderTest*"
    ANDROID_HOME=~/Library/Android/sdk ./gradlew :shared:testDebugUnitTest --tests "*DiagnosticsAttributeRegistryParityTest*"

Both pass (17 + 2 tests). No full build run.

Heads-up for anyone reproducing: without `ANDROID_HOME` set, Gradle fails with "SDK location not found" — and if you pipe the output, stale `TEST-*.xml` files from an earlier run can read as a false pass.


Quick104:
Good catch — fixed. The vendored copy now records the silo-server commit these registry additions came from (`e4627b76`), so the stated provenance matches the contents. This is exactly the class of drift the PR is meant to eliminate, so it would have been a poor thing to leave stale.

PR #249, opened 2026-08-26
fix(navigation): keep Downloads visible from launch

blurbery:
## What I changed

I noticed the Downloads tab was missing when the Android app first opened, then suddenly appeared after download records or local storage finished loading. That made the whole bottom bar reflow after launch.

I changed Downloads into a fixed mobile navigation destination from the first rendered frame. The Downloads screen still owns its empty and unavailable states, while capability and record hydration no longer decide whether the navigation item exists.

This also removes the startup record/local-byte scan that was only being used to decide tab visibility. Offline launch routing is unchanged.

## Testing

I designed and tested this behavior to make sure the bottom bar stays stable from launch.

- Android navigation unit tests pass
- Verified Downloads is present in the fixed tab list before records load
- Verified the existing offline Downloads start logic remains covered

## AI disclosure

I designed the navigation behavior and tested the result. I used AI assistance to help implement the change, simplify the old visibility path, and add regression coverage.

### DIFF
--- a/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/navigation/MobileDownloadVisibility.kt
+++ b/androidApp/src/androidMain/kotlin/org/siloserver/silo/android/ui/navigation/MobileDownloadVisibility.kt
@@ -18,11 +18,6 @@ fun hasLocalDownloadsForScope(
     profileId: String?,
 ): Boolean = scopedLocalDownloadBytes(storage, serverId, profileId) > 0L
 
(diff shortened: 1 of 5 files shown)


blurbery:
@coderabbitai full review

blurbery:
Closing this because the matching Apple Downloads PR #198 has been superseded by the finalized iOS work in Silo-Server/silo-apple#212. I am withdrawing this separate Android counterpart as part of closing the earlier paired contributions rather than leaving an orphaned cross-platform PR.

PR #250, opened 2026-08-27
fix(playback): break output-route replan loops

blurbery:
## What I changed

I split Android audio output state into two concepts: full runtime state for diagnostics and a stable planning identity for deciding whether playback must be replanned. The planning identity contains the physical sink and HDR, passthrough, and channel capabilities that can change the server recipe. It excludes Spatializer state because the server does not route on that value and Media3 can change it as a consequence of remounting.

The output-route generation now advances only when that planning identity changes. I also retain attempted-plan history when an opaque context ID changes without a material route change, so a callback cannot reopen direct, remux, and transcode routes that already failed.

## Why

I traced a live Android session that repeatedly cycled through direct play, remux, and transcode. The server transports were healthy. Media3 remounts triggered Spatializer callbacks, each callback was treated as a new physical output route, and the resulting replan cleared failed-route history. Loop detection therefore never got a stable history to act on.

This removes the false route change at its source. The companion server guard in [Silo-Server/silo-server#789](https://github.com/Silo-Server/silo-server/pull/789) protects already-installed builds that still send the legacy invalidation.

This change addresses route churn only. It does not claim to make an unsupported E-AC-3 Bluetooth path direct-play; the separate server fallback preserves the original video and converts only audio when Android genuinely reports a decoder failure.

## Validation

- Android Builds / Unit tests passed on the production-fork PR.
- Android Builds / Lint passed on the production-fork PR.
- Focused tests cover stable planning identity, real capability changes, and preserving fallback history.
- Local Gradle execution was unavailable because the validation Mac had no Java runtime; GitHub Actions supplied the repository build environment.

Source implementation and production validation: [blurbery/silo-android#4](https://github.com/blurbery/silo-android/pull/4).

Related issue: N/A — narrow fix reproduced on a live deployment.

## Scope and risk

A real sink, HDR, passthrough, or channel-capability change still advances route generation and replans normally. Only state that cannot affect the server playback recipe is filtered out. No API, schema, configuration, or Android TV product-surface changes are included.

## AI Disclosure

- Tool(s): OpenAI Codex desktop
- Model(s): gpt-5.6-sol
- Involvement: AI-assisted; I reproduced the live loop, designed the planning-identity boundary, reviewed the implementation, and validated the deployed server/client interaction
- Adversarial review: Traced the callback-to-replan path and checked each excluded field against the server planner. Focused tests verify that Spatializer-only drift is ignored while real sink, HDR, passthrough, and channel changes still invalidate the route, and attempted fallback history survives an opaque-token-only change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio route change detection by distinguishing meaningful physical or capability changes from spatializer and generation-only updates.
  * Prevented unnecessary playback replanning when output changes do not affect the active route.
  * Preserved fallback history for non-material route updates while restarting it for meaningful output changes.

* **Tests**
  * Added coverage for route identity changes, spatializer updates, device capabilities, and fallback-history behavior.

### DIFF
--- a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/AudioCapabilityManager.kt
+++ b/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/AudioCapabilityManager.kt
@@ -39,6 +39,31 @@ data class AudioPlaybackRouteSnapshot(
 )
 
 /**
+ * Identity used to decide whether playback must be replanned for a new output.
+ *
+ * Spatializer enablement is deliberately excluded. Android can toggle it while
+ * Media3 tears down and recreates an audio sink, so including it makes a player
+ * remount look like a physical route change and feeds the remount back into
+ * another server replan. The server does not route on this flag; it remains in
+ * the published diagnostics/capability snapshot.
+ */
+internal data class AudioPlanningRouteIdentity(
+    val sinkType: String,
+    val routeHashes: List<String>,
+    val capabilities: AudioPassthroughCapabilities,
+)
+
+internal fun audioPlanningRouteIdentity(
+    sinkType: String,
+    routeHashes: List<String>,
+    capabilities: AudioPassthroughCapabilities,
+): AudioPlanningRouteIdentity = AudioPlanningRouteIdentity(
+    sinkType = sinkType,
+    routeHashes = routeHashes.distinct().sorted(),
+    capabilities = capabilities.copy(spatializerEnabled = false),
+)
+
+/**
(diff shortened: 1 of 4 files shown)


blurbery:
Related server playback work: [Silo-Server/silo-server#789](https://github.com/Silo-Server/silo-server/pull/789) guards already-installed clients against false route invalidation, while [Silo-Server/silo-server#790](https://github.com/Silo-Server/silo-server/pull/790) preserves copied video for genuine Android E-AC-3 Bluetooth decoder failures.

PR #259, opened 2026-08-29
fix(tv): prefer native mounted transport for quick skips instead of re-anchoring + map seek input + fix second press for seek

evulhotdog:
## Problem

Android TV: pressing D-pad left/right during playback shows the
buffering spinner and forces a server round trip that restreams the content,
while the remote's rewind/fast-forward buttons jump instantly using locally
mounted content.

### TL;DR and what you'll notice as a user:

 - D-pad left/right skips are instant. Within the content the player already has, skips no longer show the buffer screen or hit the server — they're plain local seeks (usually measured @ ~0.1–0.5 s).
   Previously every skip re-requested the stream from the server.
 - The rewind/fast-forward buttons work on the first press. Previously the first press just pulled up the controls and you had to press again.
 - Those buttons now skip the same amounts as the on-screen buttons (−10 s / +30 s) instead of Media3's ±5/±15 s defaults.
 - In Watch Together rooms, the rewind/FF buttons no longer desync the room and they now respect the same gating as everything else.

Two cooperating causes:

1. For growing-HLS transcodes and copy-remux transports, the server publishes
   an open-ended seek window with `can_seek_anywhere=false` (it cannot keep the
   produced head current, and web clients would 404 past it — see the growing-
   manifest branch and `configureCopyRemuxTimelineV3` in `playback_v3.go`).
   The client's `decideSeek` therefore returned `ServerReanchor` for every
   seek on those transports: buffering UI, a protocol-V3 `seek_reanchor` round
   trip, and a remount — even when the target sat inside content the player
   already had mounted.
2. `KEYCODE_MEDIA_REWIND` / `KEYCODE_MEDIA_FAST_FORWARD` were unmapped, so the
   player bridge's null-action branch only revealed the transport controls and
   swallowed the first press (users had to press twice to seek), and the
   second press fell through to Media3's session-level `seekBack`/
   `seekForward`, bypassing the route-aware seek path and the Watch Together
   room transport gate.

## Approach

Three commits; the second and third are separable if you prefer them
standalone.

1. `33feccdf` — `decideSeek` gains an optional `mountedSeekableSourceRange`:
   the source-time extent the currently mounted transport provably covers.
   The TV screen reports the Media3 window's seekability and length alongside
   its existing 500 ms position poll; a target inside that range is a plain
   `Player.seekTo`. The hint is only claimed when the window is seekable with
   a known length, which covers append-only HLS manifests (window end =
   produced head), completed transcodes, and indexed streams; a growing
   progressive copy remux has no known length and stays excluded. Targets the
   mounted transport cannot serve — before the stream origin, past the
   produced head — still reanchor, and the existing same-route reanchor
   recovery (404/410/416) remains the fallback for native seeks that fail.
   The parameter is optional, so the audiobook and Cast call sites are
   unchanged.
2. `dc234833` — map REWIND/FAST_FORWARD/SKIP_BACKWARD/SKIP_FORWARD to
   `SkipBack`/`SkipForward`. They ignore `dpadHorizontalSeek` because they
   are never focus navigation, and UP halves plus auto-repeats are consumed
   so the system media-key fallback cannot seek a second time.
3. `328d12b3` — invalidate the mounted-window facts when the transport is
   replaced. The facts describe exactly one Media3 item: reports are dropped
   while the transport mount gate suppresses position reports (the same guard
   `onPositionChanged` applies), and a won mount clears them before the
   queued after-mount seek is re-evaluated. Without this, the previous
   transport's window end maps through the new plan's timeline offset and can
   overstate the new extent — a target the new transport cannot serve becomes
   a silent, wrong-position native seek with no error and no self-heal. Until
   the next poll tick (≤500 ms) the hint is simply absent, which at worst
   costs an unnecessary reanchor.

Behavioral changes, end to end:

- Quick skips within the mounted transport's proven extent no longer hit the
  server: no buffering screen, no `seek_reanchor` call, no remount. Skips the
  mounted transport cannot serve reanchor exactly as before.
- The remote's rewind/fast-forward keys seek on the first press, take the
  same route-aware path as D-pad, and skip −10 s/+30 s matching the on-screen
  transport labels instead of Media3's default ±5/±15 s session increments.
- In Watch Together rooms, media-key seeks now pass the room transport gate
  instead of forking playback.
- After a remount (reanchor, subtitle/audio replan, content switch), seeks
  commit only once the new transport's window is known; the handoff window
  degrades to a reanchor rather than a guess.

Alternatives considered and rejected: flipping `can_seek_anywhere` server-side
for event playlists (web clients seek past the produced head and 404); waiting
for the manifest head to grow on short overshoots (stall/latency risk for
marginal gain — reanchor already settles in about a second).

## Validation

- `./gradlew :android-shared:testDebugUnitTest --tests "org.siloserver.silo.common.player.seek.PlaybackTimelineSeekPolicyTest" :androidTvApp:testDebugUnitTest --tests
"org.siloserver.silo.tv.ui.screens.player.TvPlayerRemoteKeyActionTest" --tests "org.siloserver.silo.tv.ui.screens.player.TvPlayerWindowFactInvalidationTest"`
  → BUILD SUCCESSFUL. Adds 3 policy tests (hint allows native inside an open
  window; hint does not cover targets beyond the proved extent; hint cannot
  rescue targets outside the published window), 2 key-mapping tests (first
  press skips, repeat/UP consumed; media keys skip even when the idle overlay
  owns focus), and 2 invariant tests locking the mounted-window-fact
  invalidation (handoff suppression; won-mount reset ordered before the
  queued-seek re-evaluation).
- `./gradlew :androidTvApp:compileDebugKotlin :android-shared:compileDebugUnitTestKotlin`
  → success.
- `./gradlew :androidTvApp:assembleDebug`, installed on a Shield over ADB and
  exercised by hand on a growing-HLS episode (build at
  `33feccdf`+`dc234833`). 48 seeks in the logcat trace: 32 committed
  `action=native` with a non-null `mounted_source_range` (settle 85–500 ms,
  no server interaction); 16 reanchor-class commits, only for targets before
  the stream origin (`OutsideSeekWindow`) or past the produced head
  (`UnknownSeekWindow`), each settling in ~0.8–1.3 s; 0 `PlaybackException`s.
  The final build (`328d12b3`) additionally invalidates the window facts
  across remounts — that path degrades to the reanchor behavior this trace
  already exercises. Representative excerpt (raw):

  ```
  I/TvPlayerViewModel: seek_commit seek_id=23 action=native target_source_seconds=415.921 target_player_seconds=38.044 mounted_source_range=377.877..435.935
  I/TvPlayerViewModel: seek_settled seek_id=23 target_source_seconds=415.921 actual_source_seconds=416.117
  I/TvPlayerViewModel: seek_commit seek_id=32 action=server_reanchor target_source_seconds=370.427 reason=OutsideSeekWindow
  I/TvPlayerViewModel: seek_settled seek_id=32 target_source_seconds=370.427 actual_source_seconds=370.426
  ```

## Summary by CodeRabbit

- **New Features**
  - TV media rewind, skip, and fast-forward keys now support single-press seeking across player states.
  - Seeking within already loaded, seekable content now responds more quickly without unnecessary server re-anchoring.

- **Bug Fixes**
  - Prevented stale playback-window information from affecting seeks during transport changes.
  - Improved handling of seek targets outside the currently available playback range.
  - Auto-repeat and key-release events no longer trigger duplicate seek actions.

### DIFF
--- a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/seek/PlaybackTimelineSeekPolicy.kt
+++ b/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/seek/PlaybackTimelineSeekPolicy.kt
@@ -143,9 +143,19 @@ fun PlaybackTimeline.seekRestorationMode(): PlaybackSeekRestoration = when (seek
  * [PlaybackTimeline.canSeekAnywhere] is false. With no complete window, the
  * server's explicit `can_seek_anywhere` claim is required; a false claim plus
  * an absent or partial window is deliberately treated as unknown and reanchored.
- * Known bounds remain authoritative even when `can_seek_anywhere` is true.
+ * Known bounds remain authoritative even when `canSeekAnywhere` is true.
+ *
+ * Exception: the caller may prove local seekability itself by passing
+ * [mountedSeekableSourceRange] — the source-time extent the currently mounted
+ * transport provably covers (e.g. an append-only HLS manifest's produced head,
+ * or a stream the player reports as seekable with a known length). A target
+ * inside that extent is a plain `Player.seekTo` even when the published window
+ * is open-ended; only targets the mounted transport cannot serve reanchor.
  */
-fun PlaybackTimeline.decideSeek(targetSourcePositionSeconds: Double): PlaybackSeekDecision {
+fun PlaybackTimeline.decideSeek(
+    targetSourcePositionSeconds: Double,
+    mountedSeekableSourceRange: ClosedRange<Double>? = null,
+): PlaybackSeekDecision {
     require(targetSourcePositionSeconds.isFinite() && targetSourcePositionSeconds >= 0.0) {
         "Seek target must be a finite, non-negative source position."
     }
@@ -175,11 +185,13 @@ fun PlaybackTimeline.decideSeek(targetSourcePositionSeconds: Double): PlaybackSe
         )
     }
     if (!window.complete && !canSeekAnywhere) {
-        return PlaybackSeekDecision.ServerReanchor(
-            targetSourcePositionSeconds = targetSourcePositionSeconds,
-            restoration = restoration,
-            reason = ServerReanchorReason.UnknownSeekWindow,
-        )
+        if (mountedSeekableSourceRange?.contains(targetSourcePositionSeconds) != true) {
+            return PlaybackSeekDecision.ServerReanchor(
+                targetSourcePositionSeconds = targetSourcePositionSeconds,
+                restoration = restoration,
+                reason = ServerReanchorReason.UnknownSeekWindow,
+            )
+        }
     }
 
     val playerPosition = playerPositionForSource(targetSourcePositionSeconds)
--- a/android-shared/src/androidUnitTest/kotlin/org/siloserver/silo/common/player/seek/PlaybackTimelineSeekPolicyTest.kt
+++ b/android-shared/src/androidUnitTest/kotlin/org/siloserver/silo/common/player/seek/PlaybackTimelineSeekPolicyTest.kt
@@ -161,6 +161,7 @@ class PlaybackTimelineSeekPolicyTest {
         assertEquals(ServerReanchorReason.InvalidTimelineMapping, reanchor.reason)
     }
 
+    /** An open server window cannot override a proven mounted extent: targets inside it seek natively. */
     @Test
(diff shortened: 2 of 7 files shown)


evulhotdog:
@coderabbitai review

evulhotdog:
@coderabbit review

evulhotdog:
@CodeRabbit review

PR #268, opened 2026-09-01
fix(playback): direct-play Android MKV with Media3 FFmpeg

blurbery:
> [!IMPORTANT]
> **Coupled upstream change:** this Android PR must be merged together with [Silo-Server/silo-server #876](https://github.com/Silo-Server/silo-server/pull/876). Do not merge either PR independently; they form one coordinated Android playback contract change.

## Summary

Android already supports MKV. The previous player nevertheless adapted many MKVs because Android TV hid FFmpeg-only audio decoders from the server, its FFmpeg probe was not strong enough to prove native decoder availability, and later track-selection decisions could disagree with the source track the server had validated.

This PR completes the whole path on both phone and TV:

1. ship and runtime-probe the current AndroidX Media3 FFmpeg audio extension;
2. advertise only codecs the installed APK can actually execute;
3. let protocol v3 choose original MKV when the container, video, selected audio, device, and active output route are compatible;
4. make manual version/audio choices, per-file restores, quality preferences, and language settings resolve through one precedence chain; and
5. prove the selected source track after Media3 mounts it, with a bounded typed fallback when original delivery cannot honor the plan.

No playback account names, media titles, addresses, device identifiers, session identifiers, paths, or other personal activity data are included here.

## Root cause and proof from the previous player

| Finding | Immutable evidence | Playback consequence |
| --- | --- | --- |
| MKV itself was already supported | The previous client explicitly advertised `mkv` and `matroska` for original playback: [`PlaybackContainerPolicy.kt` lines 3-11](https://github.com/blurbery/silo-android/blob/2263d53e6eeec0c16297439feb23bddc0dd3100a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/video/PlaybackContainerPolicy.kt#L3-L11). | Container support was not the reason these files adapted. |
| The FFmpeg renderer was already packaged and enabled | The build included the FFmpeg AAR: [`build.gradle.kts` lines 60-68](https://github.com/blurbery/silo-android/blob/2263d53e6eeec0c16297439feb23bddc0dd3100a/android-shared/build.gradle.kts#L60-L68). Media3 registered extension renderers after platform renderers: [`SiloPlayerFactory.kt` lines 172-188](https://github.com/blurbery/silo-android/blob/2263d53e6eeec0c16297439feb23bddc0dd3100a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/SiloPlayerFactory.kt#L172-L188). | Hardware video decode could be retained while FFmpeg handled only unsupported audio; replacing Media3 was unnecessary. |
| Android TV hid FFmpeg-only codecs | The capability merge was guarded by `ffmpegAvailable && !isTv`: [`PlaybackCapabilityDetector.kt` lines 577-599](https://github.com/blurbery/silo-android/blob/2263d53e6eeec0c16297439feb23bddc0dd3100a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackCapabilityDetector.kt#L577-L599). | DTS/DTS-HD/TrueHD support existed in the installed player but was omitted from the TV payload, so the server could not authorize original playback. |
| The FFmpeg probe proved only Java class presence | The probe was only `Class.forName(...)`: [`FfmpegAudioSupport.kt` lines 71-87](https://github.com/blurbery/silo-android/blob/2263d53e6eeec0c16297439feb23bddc0dd3100a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/FfmpegAudioSupport.kt#L71-L87). | It could not distinguish a usable JNI decoder from a missing ABI, native-load failure, disabled build, or partial decoder build. |
| DTS Express and Matroska ALAC were absent from the capability table | The old binding list ended at DTS-HD: [`FfmpegAudioSupport.kt` lines 29-69](https://github.com/blurbery/silo-android/blob/2263d53e6eeec0c16297439feb23bddc0dd3100a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/FfmpegAudioSupport.kt#L29-L69). Media3 1.11 adds Matroska ALAC extraction and related DTS extraction work in the [official release notes](https://developer.android.com/jetpack/androidx/releases/media3#1.11.0). | Those tracks were under-advertised even when the native build could decode them. |
| Plain E-AC-3 decode was treated as proof of JOC decode | The previous fallback was explicit in [`PlaybackCapabilityDetector.kt` lines 769-803](https://github.com/blurbery/silo-android/blob/2263d53e6eeec0c16297439feb23bddc0dd3100a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackCapabilityDetector.kt#L769-L803). Media3 1.11 removed the comparable Pixel fallback because it was not reliable; see the [Media3 release notes](https://developer.android.com/jetpack/androidx/releases/media3#1.11.0). | Some devices could over-claim JOC and fail only after the server selected direct playback. |
| Non-default source audio lacked the server-required proof | `original_http.validated_claims` did not contain `client_selected_audio_track_v1`. | The server correctly rejected original delivery when a non-default source track was requested. |
| Automatic language and version decisions were split across layers | Audio language was applied as a Media3 preference after planning; TV returned `audioTrackIndex = 0` even when v3 selected another track; phone copied a raw audio ordinal into another file during version switching. | The server could validate one track while the player selected another codec/language, and an ordinal could become a different track after a file switch. |
| Selection recovery was not a real deadline | Phone counted three track snapshots; TV had no time-bounded confirmation failure. | Callback frequency could make recovery too early, too late, or never occur. |

The server-side causal chain is also explicit:

1. Original audio is accepted only when the active sink has an exact passthrough claim or the selected source codec appears in the client's audio-decode capabilities; otherwise the server records `audio_codec_unsupported`: [`capabilities_v3.go` lines 298-331](https://github.com/blurbery/silo-server/blob/78e171a92df27f803143c75e3a338d854f0301ec/internal/playback/capabilities_v3.go#L298-L331).
2. Original playback requires `audioOK`; when false, planning falls through to remux/HLS and may convert audio to AAC: [`plan_v3.go` lines 463-499](https://github.com/blurbery/silo-server/blob/78e171a92df27f803143c75e3a338d854f0301ec/internal/playback/plan_v3.go#L463-L499).

### Sanitized production evidence

- The supplied admin capture showed Android TV source `MKV + H.264 1080p + DTS-HD MA 5.1` delivered as `HLS + H.264 1080p + AAC 5.1`; the UI marked video and audio as adapted.
- A read-only audit requested the latest 5,000 container-log lines and the available window returned 503 entries. Fifteen MKV-related records belonged to one playback session. All 13 records exposing `target_audio_codec` selected AAC; records exposing `target_video_codec` showed 3 copy attempts and 10 H.264 encode attempts.
- This is evidence for the reported route and retry behavior, not a population-wide frequency claim.

## Existing Media3/FFmpeg changes retained in this PR

- Upgrade AndroidX Media3 and the checked-in FFmpeg extension from 1.10.1 to stable 1.11.0.
- Build the native extension from pinned Media3 1.11.0, FFmpeg n6.0, and NDK r26d inputs.
- Package `arm64-v8a`, `armeabi-v7a`, and `x86_64`, and verify every ELF LOAD segment for 16 KB page alignment.
- Check both JNI availability and each exact decoder at runtime; missing ABIs, native-load failures, disabled builds, and partial builds fail closed.
- Enable the verified FFmpeg fallback on TV as well as phone.
- Cover AC-3, E-AC-3/JOC, TrueHD, DTS core/Express/HD, and ALAC.
- Keep platform MediaCodec and encoded passthrough renderers ahead of FFmpeg. FFmpeg is a software-audio fallback; video remains on hardware MediaCodec.
- Stop treating a plain E-AC-3 platform decoder as proof of JOC support.
- Keep the existing FFmpeg build flag as an emergency kill switch.

## New direct-play and selector changes

- Scope `client_selected_audio_track_v1` to `original_http` only. It is not leaked into progressive or HLS delivery.
- Resolve the file first, including Auto/Original/4K/1080p and other quality choices.
- Resolve audio against that selected file with one phone/TV precedence chain:
  1. current manual/session choice;
  2. resolved episode carry-over;
  3. per-file persisted choice;
  4. audio-language setting/profile fallback;
  5. file default.
- Translate a manual choice across file versions by language, codec family, channel count, and title. A raw list ordinal never crosses a file boundary, and an ambiguous identity falls back to the target file's automatic chain.
- Return the server-selected audio index from the TV starter instead of silently defaulting to zero.
- Reconcile the server catalog ordinal with Media3's mounted track identity on phone and TV.
- Require mounted-identity confirmation for byte-for-byte original delivery; transcoded representations continue to confirm from the server plan because their output cannot identity-match the source row.
- Use one non-sliding 5-second deadline per selection generation. Repeated Media3 snapshots cannot postpone it and a resolved generation cannot expire.
- Emit typed `audio_track_selection_failed` recovery when the original file cannot mount and select the promised track. Protocol v3 retains the selected audio intent, skips the rejected plan key, and chooses the next packaged route.
- Preserve source audio and hardware video whenever the device/output can execute them; keep HLS/AAC/downmix as the constrained fallback for missing decoders or channel ceilings.

## Android TV boxes and output routes

This is route-driven, not model-whitelisted.

- The existing output manager watches HDMI/EDID renegotiation, device changes, and encoded-surround settings, producing a new capability snapshot when the AVR/soundbar route changes: [`AudioCapabilityManager.kt` lines 66-79](https://github.com/blurbery/silo-android/blob/2263d53e6eeec0c16297439feb23bddc0dd3100a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/AudioCapabilityManager.kt#L66-L79).
- On Android 10+, it probes encoded codecs and layouts with the platform direct-play API and sends only successful bitstream entries: [`AudioCapabilityManager.kt` lines 336-395](https://github.com/blurbery/silo-android/blob/2263d53e6eeec0c16297439feb23bddc0dd3100a/android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/AudioCapabilityManager.kt#L336-L395).
- A Shield-class device can retain its broader verified HDMI passthrough path when the receiver supports it. A Google TV Streamer-class device advertises only the narrower route it actually reports; other compiled formats may still use verified local FFmpeg decode. No passthrough capability is inferred from a model name.

### Previously completed proof retained

- **42 focused assertions passed** for runtime probing, capability advertisement, phone/TV behavior, JOC handling, DTS Express, ALAC, passthrough, and track selection.
- The prior full PR validation run, [GitHub Actions run 33510590138](https://github.com/blurbery/silo-android/actions/runs/33510590138), passed **all 1,388 unit tests** and Android lint across shared, phone, and TV modules at the previously validated head.
- Dependency locks, checksum metadata, XML validation, diff checks, and build-supply-chain checks passed.
- The reproducible native build, [GitHub Actions run 33504466713](https://github.com/blurbery/silo-android/actions/runs/33504466713), passed for all three ABIs and verified 16 KB ELF alignment.
- Verified AAR SHA-256: `bbe661491dd342f2930ffd3bafd6f3e496da3d84586cc5d603432d49daea6b4f`.

### Tests added for the new changes

- Delivery-scope tests prove `client_selected_audio_track_v1` appears only on original HTTP.
- Runtime reconciliation tests prove original delivery requires mounted identity.
- Watchdog tests prove the deadline does not slide/re-arm and a resolved selection never expires.
- Shared selection tests cover language aliases, explicit/persisted/language precedence, and semantic cross-version mapping.
- Phone and TV starter tests prove the selected source audio reaches the initial v3 request.
- Upstream server companion PR [Silo-Server/silo-server #876](https://github.com/Silo-Server/silo-server/pull/876), from `blurbery:test/android-direct-play-contract-upstream` into upstream `main`, adds the Android-shaped planner contract for original MKV, non-default audio, rejected-plan recovery, missing decoders, and channel ceilings. It is ready for review and cross-linked to this Android PR; server runtime behavior is unchanged.

The equivalent fork head, commit `0ee0b02`, passed both unit tests and Android lint in [GitHub Actions run 33517017381](https://github.com/blurbery/silo-android/actions/runs/33517017381). The clean upstream-based port, head `b7513ac1`, preserves only the intended player commit series and passed unit tests in 4m43s plus Android lint in 11m10s in [upstream GitHub Actions run 33519639644](https://github.com/Silo-Server/silo-android/actions/runs/33519639644). Local Gradle execution is unavailable on this Mac because no Java runtime is installed; remote CI is the authoritative compile/test/lint result for the upstream port.

## Risk and rollback

- Decoder claims fail closed: a codec is advertised only if the installed JNI library loads and confirms that exact MIME decoder.
- Passthrough remains output-route-specific and ahead of FFmpeg.
- Original-file selection is now verified at runtime and has a bounded fallback instead of an unbounded optimistic claim.
- Progressive remux remains intentionally disabled on Android pending a seek-safe transport; HLS is the packaged fallback.
- Reverting the latest selector/claim commit restores the previously validated Media3/FFmpeg behavior; reverting the full PR restores the old packaged-playback behavior.

## AI Disclosure

- Harness: OpenAI Codex desktop
- Tool(s): Codex local command execution and GitHub CLI
- Model(s): `gpt-5.6-sol` (reasoning effort: `ultra`)
- Involvement: AI-assisted
- Direction: User-directed. The repository owner explicitly requested the Android player scope, upstream target, cross-PR linkage, validation, and publication.
- Adversarial review: Traced the client capability, planning, source-track selection, runtime confirmation, and typed-fallback paths across phone and TV; added positive and negative coverage, then required unit-test and lint success on the published head.

No Silo server configuration, deployment, containers, environment, database, or production service was changed.

## Summary by CodeRabbit

* **New Features**
  * Updated playback components to Media3 1.11.0.
  * Added ALAC support and improved DTS Express handling.
  * Audio preferences now honor configured language, manual selections, and choices carried across file versions.
  * Improved audio-track detection and fallback behavior across phones and TVs.
  * Added verification and recovery when selected audio tracks cannot be confirmed.
  * Improved localized date formatting using the active device locale.

* **Documentation**
  * Updated playback, FFmpeg, and dependency references for Media3 1.11.0.


blurbery:
Superseded by #267 after the clean upstream-based commit series was applied to its original source branch with force-with-lease.

blurbery:
Reached Code Rabbit review limit :(

android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/video/AudioReconcile.kt
```diff
@@ -96,7 +107,7 @@ fun reconcileDesiredAudioAction(
     // remains, so a main mix and its commentary — same language, same codec —
     // would confirm each other.
     val target = matchMountedAudioTrack(wanted, mounted)
-        ?: return if (planAudioOrdinal == desired.catalogOrdinal) {
+        ?: return if (!requiresMountedIdentity && planAudioOrdinal == desired.catalogOrdinal) {
```
Quick104, line 118: **Blocking: healthy direct play gets replanned or torn down after 5 s.**

`requiresMountedIdentity` turns the null-match case into `None`, which arms the watchdog and ends in a failure-classified `audio_track_selection_failed` replan. But `matchMountedAudioTrack` returns null in routine cases on a byte-identical original mount:

- Media3's `Format` constructor normalizes every mounted language to ISO 639-1 (`Util.normalizeLanguageCode` builds its table from `Locale.getISOLanguages`), while `canonicalSubtitleLanguage` aliases only seven codes. Catalog `ita` vs mounted `it` is an empty pool.
- A mounted `audio/eac3-joc` canonicalizes to `eac3joc`, which does not equal the catalog's `eac3`.
- Two untitled tracks with the same language, codec, and layout tie.

TV seeds `desiredAudio` on every load via `primePlannedAudio`, so this is every direct-play start with such a track. Because the classification is failure-shaped, a 409 or Terminal from the replan sets a fatal error instead of the non-fatal track-change intent the base used on phone.

Fix in https://github.com/Silo-Server/silo-android/commit/18b3d6035b7261510e091e6f7cbbe32b5392277c: full ISO 639-2 table in `SubtitleLanguage.kt`, `eac3-joc` folded into the `eac3` family, and a positional fallback for original mounts when inventory sizes match and no stated field contradicts. Tests cover all three.

android-shared/src/androidMain/kotlin/org/siloserver/silo/common/player/PlaybackCapabilityDetector.kt
```diff
@@ -774,21 +787,17 @@ internal fun canDecodeAudio(
  * channels" — those are different answers for the viewer and different
  * fallbacks for the server.
  *
- * Media3 soft-matches E-AC3 JOC onto a plain E-AC3 decoder
- * (`MediaCodecUtil.getAlternativeCodecMimeType`), so a JOC track is accepted by
- * an E-AC3 decoder here too; refusing it would reject content Media3 plays.
+ * Keep E-AC3 JOC separate from plain E-AC3. Media3 1.11 removed its Pixel
```
Quick104: **The Media3 claim here is not accurate, and the change refuses content Media3 plays.**

1.11.0 did not remove the JOC-to-E-AC-3 soft match. `MediaCodecUtil.getAlternativeCodecMimeType` still returns `audio/eac3` for JOC; it is gated by `supportsEac3JocFallbackDecoding()`, which only returns false when `Build.MANUFACTURER` is `Google`. So on Shield, Sony, onn, Fire TV, and so on, Media3 still decodes JOC through the plain E-AC-3 decoder.

With FFmpeg loaded this is masked because `ffmpegSupportsMime` rescues the verdict. With the kill switch or a JNI load failure (the fail-closed path this PR designs for), a JOC track on a non-Google device now preflights as `UnsupportedAudioCodec`. The inverted test pins the wrong behavior.

https://github.com/Silo-Server/silo-android/commit/18b3d6035b7261510e091e6f7cbbe32b5392277c mirrors Media3 exactly: `platformCanDecodeAudio` takes a `jocFallsBackToEac3` flag defaulting to the same manufacturer check.

gradle/libs.versions.toml
```diff
@@ -7,7 +7,7 @@ kotlinx-serialization = "1.8.1"
 kotlinx-coroutines = "1.10.2"
 koin = "4.1.0"
 coil = "3.1.0"
-media3 = "1.10.1"
+media3 = "1.11.0"
```
Quick104, line 10: **Blocking: StackOverflowError on every HLS remux start with Media3 1.11.**

Media3 1.11 made both `AudioSink.configure` overloads default interface methods that call each other. Kotlin `by delegate` forwards only abstract members, so neither overload reaches `DefaultAudioSink` and the first audio format change recurses until the process dies:

```
java.lang.StackOverflowError: stack size 1036KB
    at androidx.media3.exoplayer.audio.AudioSink.configure(AudioSink.java:718)
    at androidx.media3.exoplayer.audio.AudioSink.configure(AudioSink.java:703)
    ...
    at androidx.media3.exoplayer.audio.MediaCodecAudioRenderer.onOutputFormatChanged(MediaCodecAudioRenderer.java:718)
```

Reproduced on the onn box; it killed the app about a second after `audio decoder initialized`. Fix is to extend `ForwardingAudioSink` instead, which forwards every member including the new defaults: https://github.com/Silo-Server/silo-android/commit/3a858529a34b5b58b5e95ff2f6c2a1b50e59b14f. The added test overflows the stack against the current code.

The sink itself (`PassthroughSuppressionRegistry.kt:76`) is unchanged by the PR, which is why this is anchored on the version bump that makes it fatal.

---

**Also from this bump: an undisclosed Compose framework upgrade** (via `media3-ui-compose`, line 117, unchanged in the diff).

`media3-ui-compose:1.11.0` imports `compose-bom:2026.05.00`, so the lockfiles move both apps from Compose UI 1.9.2 / Material3 1.3.1 to UI 1.11.1 / Material3 1.4.0, plus kotlin-stdlib 2.2.20 against a 2.1.20 compiler. Nothing in either app references `androidx.media3.ui.compose`, and three of the eleven commits (`fix(ci)` calendar locale, TV Compose lint, pill requesters) are fallout from that bump rather than from the playback change.

It also breaks the invariant on `compose-ui-test = "1.9.2"` (line 30): the TV unit-test runtime classpath stays on `ui:1.9.2` while the app ships `ui:1.11.1`.

https://github.com/Silo-Server/silo-android/commit/18b3d6035b7261510e091e6f7cbbe32b5392277c drops the unused dependency and regenerates locks and verification metadata; both apps go back to 1.9.2 / 1.3.1. If you do want the Compose upgrade, it should be its own PR with UI screenshots.

Quick104 (changes requested):
Tested the head (`b7513ac1`) on an onn 4K box (armeabi-v7a, Android 14) against shared dev, plus a static pass over the diff. Two blocking defects and one dependency concern. Fixes for all of them are on `review/pr-268-android-fixes` as two commits you can cherry-pick:

- https://github.com/Silo-Server/silo-android/commit/18b3d6035b7261510e091e6f7cbbe32b5392277c — review findings 1 to 3 below
- https://github.com/Silo-Server/silo-android/commit/3a858529a34b5b58b5e95ff2f6c2a1b50e59b14f — the StackOverflow crash (finding 4)

Full Android suite passes with both applied (7331 tests), and the box plays a 4K HEVC DV DTS-HD MKV from Start Over via `server_remux_hls` with no errors.

**Unrelated server bug you will hit while testing:** copy-video fMP4 starts from position zero wrote a negative audio `tfdt` and Media3 rejected the first segment with `Top bit not zero: -1024` on both 1.10.1 and 1.11.0. That is fixed server-side in Silo-Server/silo-server#897 and is already on shared dev.

Not blocking, but worth a sentence in the PR body: the TV change re-advertises FFmpeg-only codecs that commit `86042956` deliberately hid to keep Shield-class boxes on the tunneled A/V-sync path. No Shield evidence is in the diff.

PR #305, opened 2026-09-14
fix(tv): end the playback session when the player stops

Quick104:
## Summary

- Android TV and Fire TV devices can stop delivering lifecycle callbacks to the player while the process stays alive. Pausing alone left the 10s progress reporter heartbeating a server session that nobody was watching, so the server kept the session open indefinitely (reported on Discord).
- `TvPlayerScreen` now treats `Lifecycle.Event.ON_STOP` as a real exit: it sets `exitRequested`, calls `viewModel.stopSessionForExitAsync(...)` with the controller's current position and duration, and stops the `mediaController`.
- The `exitRequested` guard makes this idempotent with the existing explicit exit path, so leaving the player the normal way does not tear the session down twice.
- PiP is unaffected: the activity stays started while in picture-in-picture, so `ON_STOP` does not fire and playback keeps running.
- One file changed: `androidTvApp/src/androidMain/kotlin/org/siloserver/silo/tv/ui/screens/player/TvPlayerScreen.kt` (+15).

## Testing

- `./gradlew :androidTvApp:assembleDebug` — Not run.
- `./gradlew test` — Not run.
- Manual: background the TV player with Home and confirm the server session ends and progress stops reporting — Not run.
- Manual: enter PiP and confirm playback continues and the session stays open — Not run.
- Manual: exit the player with Back and confirm the session ends once with the correct resume position — Not run.

## Risks and follow-ups

- The position and duration come from `mediaController?.currentPosition` / `duration`, which are nullable at `ON_STOP`. If the controller is already released, the session ends without a final position and resume falls back to the last reported progress.
- Worth confirming PiP behavior on a real Fire TV device, since the fix depends on `ON_STOP` not firing there.
- No issue link available; this came in through Discord.

## AI disclosure

Written by Claude Opus 5 (`claude-opus-5`) in the Claude Code agent harness. No other AI tooling was used.

> [!NOTE]
> ### End TV playback session on player `ON_STOP` and guard detached stop
> - Adds an `ON_STOP` lifecycle branch to `TvPlayerScreen` that calls `stopPlaybackAndExit` when `exitRequested` is false, so the player leaves its destination on stop rather than relying on prior lifecycle handling.
> - Changes the `TvPlayerViewModel` exit teardown to call `stopDetached` only when `exitSessionId` is non-null via a null-safe `let` block, instead of invoking it unconditionally.
> - Behavioral Change: `stopDetached` is no longer called during exit teardown when `exitSessionId` is null; review `TvPlayerViewModel` exit teardown to confirm no session needs stopping in that case.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f217305.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - TV playback sessions now close reliably when leaving the player screen or moving the app to the background.
  - Playback position is preserved during session cleanup to support more accurate resume behavior.
  - Media controls are paused, stopped, and cleared cleanly to prevent unexpected playback.
  - Users are returned through the standard exit flow after playback cleanup completes.


Quick104:
Addressed the lifecycle review finding in commits `1ba0a6db` and `2ff55111`. When the player activity reaches `ON_STOP`, it now uses the existing `stopPlaybackAndExit()` path, which tears down the session and leaves the player destination. This prevents returning to a stopped/black route while retaining the PiP exemption. Validation: `./gradlew :androidTvApp:compileDebugKotlin` passed.

Quick104:
Fixed the teardown ownership finding in commit `$(git rev-parse --short HEAD)`. `stopSessionForExitAsync()` now skips `stopDetached` when `exitSessionId` is null, preventing an old loading route from stopping a replacement session adopted before the queued teardown runs. Validation: `./gradlew :androidTvApp:compileDebugKotlin` passed.

